PR #8, opened 2019-08-02
Add missing mixin

gcandal:
The `oauthMixin` uses an `handleError` function that is defined in the `partMixin`. Currently it crashes.


gcandal:
Already done.

PR #13, opened 2019-10-21
Expose filter changeItem

gcandal:
`filter-platforme` manages the pagination of items based on a given filter. This is problematic when we want to update a single, such as when deleting an item or changing its state.

Exposing a method allows users to apply changes without having to refresh the whole listing or losing the visual feedback.

### DIFF
--- a/vue/components/ui/filter-platforme/filter-platforme.vue
+++ b/vue/components/ui/filter-platforme/filter-platforme.vue
@@ -180,6 +180,13 @@ export const FilterPlatforme = {
         updateQuery(options) {
             const { sort, reverse, filter } = options;
             this.$router.replace({ query: { ...this.$route.query, sort, reverse, filter } });
+        },
+        changeItem(index, item) {
+            if (item) {
+                this.items.$set(index, item);
+            } else {
+                this.items.splice(index, 1);
+            }
         }
     }
 };


vue/components/ui/filter-platforme/filter-platforme.vue
```diff
@@ -180,6 +180,13 @@ export const FilterPlatforme = {
         updateQuery(options) {
             const { sort, reverse, filter } = options;
             this.$router.replace({ query: { ...this.$route.query, sort, reverse, filter } });
+        },
+        changeItem(index, item) {
```
joamag: it may be better to use `setItem` and `removeItem`

vue/components/ui/filter-platforme/filter-platforme.vue
```diff
@@ -181,12 +181,11 @@ export const FilterPlatforme = {
...
             this.$router.replace({ query: { ...this.$route.query, sort, reverse, filter } });
         },
-        changeItem(index, item) {
-            if (item) {
-                this.items.$set(index, item);
-            } else {
-                this.items.splice(index, 1);
-            }
+        setItem(index, item) {
+            if (item) this.items.$set(index, item);
```
joamag: if no item is passed then an error must occur... We should not fail silently.

joamag:
Merging this as an internal thing. If an item is removed the list should be refreshed. BTW it should not be possible to delete things from a list.

PR #14, opened 2019-10-22
Fix scroll load more

gcandal:


vue/mixins/scroll.js
```diff
@@ -40,7 +40,7 @@ const scrollMixin = {
             if (bottomScroll) {
                 return;
             }
-            this.$refs.jobFilter.loadMore();
```
joamag: the underlying layer should implement the get filter...

PR #21, opened 2019-11-15
Fixed NavigationDuplicated exception

NFSS10:
| - | - |
| --- | --- |
| Issue | - |
| Decisions | When reloading the same page an "NavigationDuplication" error was thrown. This occurs when the final location is the same as the current one. The reason the error just showed recently is because of a change in "vue-router" (https://github.com/vuejs/vue-router/issues/2872#issuecomment-519073998). This change was introduced in [v3.1.0](https://github.com/vuejs/vue-router/releases/tag/v3.1.0). <br><br>To fix the problem, we check the query and see if it changed to avoid loading to the same location.|
| Animated GIF | Below |

Before, showing the "NavigationDuplicated" error:
exception_reload_page

After, error gone:
no_exception


joamag:
@gcandal Need to think this on a more generic way

PR #36, opened 2019-11-21
Device change mixin

gcandal:

### DIFF
--- a/vue/mixins/device.js
+++ b/vue/mixins/device.js
@@ -1,0 +1,49 @@
+const DESKTOP_CLASS = "desktop";
+const TABEL_CLASS = "tablet";
+const MOBILE_CLASS = "mobile";
+
+const TABLET_WIDTH = 768;
+const MOBILE_WIDTH = 420;
+
+const deviceMixin = {
+    methods: {
+        setTabletWidth(width, update = false) {
+            this.tabletWidth = width;
+            if (update) this.updateDeviceClass();
+        },
+        setMobileWidth(width, update = false) {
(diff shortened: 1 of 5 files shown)


vue/mixins/device.js
```diff
@@ -0,0 +1,49 @@
...
+            if (width > tabletWidth) {
+                body.classList.add(DESKTOP_CLASS);
+            } else if (width > mobileWidth) {
+                body.classList.add(TABEL_CLASS);
+            } else {
+                body.classList.add(MOBILE_CLASS);
+            }
+        },
+        _getBody() {
+            if (this._body) return this._body;
```
joamag: You know that `document.body` does the job right ;)?

vue/mixins/device.js
```diff
@@ -0,0 +1,49 @@
...
+        },
+        setMobileWidth(width, update = false) {
+            this.mobileWidth = width;
+            if (update) this.updateDeviceClass();
+        },
+        listenDeviceChange() {
+            window.addEventListener("resize", () => this.updateDeviceClass());
+        },
+        updateDeviceClass() {
+            const width = window.innerWidth;
```
joamag: I've seen this code elsewhere... where was it?

gcandal: https://github.com/ripe-tech/ripe-commons-pluginus/blob/master/js/util/device.js

vue/mixins/device.js
gcandal: (facepalm)

PR #39, opened 2019-11-25
Add default height to label

gcandal:
Avoids unwanted height inheritance.


vue/components/ui/tag-platforme/tag-platforme.vue
```diff
@@ -12,6 +12,7 @@
     display: inline-block;
     font-size: 12px;
     font-weight: 600;
+    height: auto;
```
joamag: not a good idea let's force line-height instead and not auto a specific value like 14px

PR #41, opened 2019-11-25
Prevent link style inside tables

gcandal:


joamag:
@gcandal why is this?

gcandal:
Will solve this by preventing the components from inheriting this.

PR #46, opened 2019-11-27
Adding subtle mode in tags

Benedito0:
| - | - |
| --- | --- |
| Issue |  |
| Decisions |   |
| Animated GIF |tdF2qTMQJX|


gcandal:
Please do this without opacity, use pure colors instead.

Also, place those colors under `variables.css`. For instance:

```
.tag {
    color: $white;
}

.tag.blue {
    background-color: $blue;
}

.subtle-tag.blue {
    color: $blue;
    background-color: $light-blue;
}
```

PR #101, opened 2019-12-13
Select component

gcandal:


NFSS10:
Fixed some stuff.

NFSS10:
Fixed dropdown not opening when using keyboard.

PR #102, opened 2019-12-13
Add description component

3rdvision:
| - | - |
| --- | --- |
| Issue | - |
| Decisions | - |
| Animated GIF | imagem  |


gcandal:
My bad here, but I had misunderstood a talk I had with Jorge which mislead me into thinking this should be a component.

For now, every time that we need something like this, we'll implement it at the application-level. In any case, it won't happen in the near future and if it does we just need to copy it from this PR.

PR #104, opened 2019-12-16
Added details "globalEvents" prop

NFSS10:
Allows to ignore the keyboard events that currently changes the orders to the next one, or the previous one.


gcandal:
This shouldn't be needed, you just need to catch the event in the text are and stop its propagation (eg: `key.down.stop`)

PR #108, opened 2019-12-18
Remove line-height from header

BeeMargarida:
| - | - |
| --- | --- |
| Issue | -- |
| Decisions | Removed line-height from header and header-container. The change didn't have any impact in the design and was causing changes in the announcements-modal. |
| Animated GIF | -- |


joamag:
We should keep the line height, proper overriding should be done by specialization

PR #109, opened 2019-12-18
Fixed alert bug + Allow to disable keyboard navigation in "details"

NFSS10:


gcandal:
The title says it all: separate these 2 in two different PRs. Two different scopes = two different PRs. It's even more glaring because one is an improvement and the other a bug fix.

PR #123, opened 2019-12-23
Side Component Story and id+text into value+label

BeeMargarida:
| - | - |
| --- | --- |
| Issue | -- |
| Decisions | Changed side component id and text attributes to label and value, to be in agreement with other components (checkbox, radio, dropdown, ...). Also added story for side component. |
| Animated GIF | -- |

### DIFF
--- a/vue/components/ui/atoms/side/side.stories.js
+++ b/vue/components/ui/atoms/side/side.stories.js
@@ -1,0 +1,65 @@
+import { storiesOf } from "@storybook/vue";
+import { withKnobs, boolean, number, select } from "@storybook/addon-knobs";
+
+storiesOf("atoms", module)
+    .addDecorator(withKnobs)
+    .add("Side", () => ({
+        props: {
+            position: {
+                default: select(
+                    "Position",
+                    {
+                        Left: "left",
+                        Right: "right"
+                    },
+                    "left"
+                )
+            },
+            width: {
+                default: number("Width", 280)
+            },
+            visible: {
+                default: boolean("Visible", true)
+            },
+            links: {
+                default: () => [
+                    {
+                        value: "jobs",
+                        label: "Jobs",
+                        link: { name: "jobs" },
+                        selected: true
+                    },
+                    {
+                        value: "job-create",
+                        label: "Create Job",
+                        link: { name: "job-create" },
+                        selected: false
+                    },
+                    {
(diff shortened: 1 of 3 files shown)


vue/components/ui/atoms/side/side.stories.js
```diff
@@ -0,0 +1,65 @@
...
+                ]
+            }
+        },
+        template: `
+            <div>
+                <side 
+                    v-bind:links="links" 
+                    v-bind:width="width" 
+                    v-bind:visible="visible" 
+                    v-bind:position='position'
```
gcandal: "position"

vue/components/ui/atoms/side/side.stories.js
```diff
@@ -0,0 +1,65 @@
...
+            }
+        },
+        template: `
+            <div>
+                <side 
+                    v-bind:links="links" 
+                    v-bind:width="width" 
+                    v-bind:visible="visible" 
+                    v-bind:position='position' 
+                    v-bind:global='true'
```
gcandal: No need for this

vue/components/ui/atoms/side/side.stories.js
```diff
@@ -0,0 +1,65 @@
...
+                        link: { name: "devices" },
+                        selected: false
+                    },
+                    { value: "separator" },
+                    { value: "settings", label: "Settings", disabled: true }
+                ]
+            }
+        },
+        template: `
+            <div>
```
gcandal: You can delete the div, I guess. That's only used when we have more than 1 HTML element in the story

vue/components/ui/atoms/side/side.stories.js
```diff
@@ -0,0 +1,65 @@
...
+                <side 
+                    v-bind:links="links" 
+                    v-bind:width="width" 
+                    v-bind:visible="visible" 
+                    v-bind:position='position' 
+                    v-bind:global='true'
+                ></side>
```
gcandal: Does `<side />` work?

vue/components/ui/atoms/side/side.stories.js
gcandal: Nevermind, let the `div` stay here and add `p` that prints the value of `visible`. Ask @3rdvision to try and explain how we're using the `.sync` for that matter

PR #133, opened 2020-01-02
Fixed details slots example

NFSS10:
Slot name expects an id.


gcandal:
Probably a `v-if="item.id"` should be added here: https://github.com/ripe-tech/ripe-components-vue/blob/55f46e74a047105586fffd0380d36a72a4d1d901/vue/components/ui/organisms/details/details.vue#L87

So that the ID becomes mandatory for all cases. It seems more coherent.

joamag:
@gcandal we need to go with the `value` instead of the `id`

PR #151, opened 2020-01-06
Missing z-index in the select

FlaviooLima:
| - | - |
| --- | --- |
| Issue | Missing z-index in the component ending up with components on top of this one. |
| Decisions |   Added z-index property |
| Animated GIF | Screenshot 2020-01-06 at 12 26 02 |

### DIFF
--- a/vue/components/ui/atoms/dropdown/dropdown.vue
+++ b/vue/components/ui/atoms/dropdown/dropdown.vue
@@ -56,6 +56,7 @@
 .dropdown-container {
     position: relative;
     width: 100%;
+    z-index: 1;
 }
 
 .dropdown {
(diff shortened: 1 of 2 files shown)


gcandal:
@FlaviooLima add it to the dropdown instead :)

FlaviooLima:
@gcandal give it a look now pls :)

gcandal:
> @gcandal give it a look now pls :)

What I meant was adding the `z-index` here: https://github.com/ripe-tech/ripe-components-vue/blob/4d298cbae8c8fe4544432c6a9b250195b19fd99e/vue/components/ui/atoms/dropdown/dropdown.vue#L61

The rationale is: regardless of where you use the drop-down, even if outside the `select-platforme`, it will always work properly.

PR #153, opened 2020-01-06
Fixed attachments names conflict warning

NFSS10:
| - | - |
| --- | --- |
| Issue | - |
| Decisions | No more warnings relative to the conflict between attachments with the same name. |
| Animated GIF |   |


vue/components/ui/molecules/chat-message/chat-message.vue
```diff
@@ -138,7 +138,7 @@ export const ChatMessage = {
         },
         message: {
             type: String,
-            required: true
+            default: null
```
gcandal: Revert, let's force clients to use the empty string for this use case.

PR #162, opened 2020-01-07
Animated Button Icon

BeeMargarida:
| - | - |
| --- | --- |
| Issue | -- |
| Decisions | Animated button icon that changes icon when clicked, which is useful for giving feedback to actions such as copying links. |
| Animated GIF | Jan-07-2020 15-54-58 |

### DIFF
--- a/vue/components/ui/molecules/animated-button-icon/animated-button-icon.vue
+++ b/vue/components/ui/molecules/animated-button-icon/animated-button-icon.vue
@@ -1,12 +1,11 @@
 <template>
-    <div class="animated-button-icon">
+    <div class="animated-button-icon" v-bind:class="{ 'not-animating': !animating }">
         <button-icon
             class="button-icon-front"
             v-bind:icon="icon"
             v-bind:color="color"
             v-bind:size="size"
             v-bind:disabled="disabled"
-            v-bind:class="{ visible: !animating }"
             v-on:click="onClick"
         />
         <button-icon
@@ -14,7 +13,6 @@
             v-bind:icon="animationIcon"
             v-bind:color="color"
             v-bind:size="size"
-            v-bind:class="{ visible: animating }"
         />
     </div>
 </template>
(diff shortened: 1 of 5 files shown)


vue/components/ui/molecules/animated-button-icon/animated-button-icon.vue
```diff
@@ -0,0 +1,106 @@
...
+            class="button button-icon-front"
+            v-bind:icon="icon"
+            v-bind:color="color"
+            v-bind:size="size"
+            v-bind:disabled="disabled"
+            v-bind:class="{ visible: !clicked }"
+            v-on:click="onClick"
+        />
+        <button-icon
+            class="button button-icon-click"
```
gcandal: button-icon-back? (as opposed to front)

I understand that for you click = animation, but I don't think that's clear for everyone reading this.

vue/components/ui/molecules/animated-button-icon/animated-button-icon.vue
```diff
@@ -0,0 +1,106 @@
...
+            v-bind:icon="clickIcon"
+            v-bind:color="color"
+            v-bind:size="size"
+            v-bind:disabled="true"
```
gcandal: Are you sure we want this? I think it changes the UI, and we don't want that. Maybe disabling pointer events is enough.

vue/components/ui/molecules/animated-button-icon/animated-button-icon.vue
```diff
@@ -0,0 +1,106 @@
...
+    left: 0px;
+    opacity: 0;
+    pointer-events: none;
+    position: absolute;
+    top: 0px;
+}
+
+.button.button-icon-click.visible {
+    opacity: 1;
+}
```
gcandal: Would you please refactor this to stop using the `button` class? My rationale is that the properties between the 2 buttons are always opposed to each other, and per the current way this is written that's not clear and takes a bit of juggling to realize what's going on.

The `.button.button-icon-front.visible` is a sympton of this, where you are overriding a property that shouldn't ever have been set in the first place.

PR #164, opened 2020-01-07
Added delete button to "search" component

NFSS10:
| - | - |
| --- | --- |
| Issue |- |
| Decisions |  Added delete button that appears when there is text in the search bar. When pressed, all that text is deleted. |
| Animated GIF | search_delete_button |

### DIFF
--- a/vue/components/ui/molecules/search/search.vue
+++ b/vue/components/ui/molecules/search/search.vue
@@ -203,6 +203,9 @@ export const Search = {
         };
     },
     watch: {
+        value(value) {
+            this.valueData = value;
+        },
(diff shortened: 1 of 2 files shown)


vue/components/ui/molecules/search/search.vue
```diff
@@ -36,6 +36,14 @@
             v-on:focus="focused = true"
             v-on:blur="focused = false"
         />
+        <slot name="icon-delete" v-if="value">
```
gcandal: No need for a slot, let's just guard this behind a flag (`enableDelete` or something) that is disabled by default.

PR #168, opened 2020-01-09
Input Symbol

3rdvision:
| - | - |
| --- | --- |
| Issue |- |
| Decisions |   |
| Animated GIF |  imagem  |

### DIFF
--- a/vue/components/ui/organisms/input-symbol/input-symbol.vue
+++ b/vue/components/ui/organisms/input-symbol/input-symbol.vue
@@ -1,10 +1,14 @@
 <template>
     <div
         class="input-symbol"
-        v-bind:class="[{ disabled: disabled }, `color-${variant}`, `border-${border}`]"
+        v-bind:class="[
+            { disabled: disabled },
+            `color-${variant}`,
+            `border-${border}`,
+            `text-align-${align}`
+        ]"
     >
         <input-ripe
-            class="input-ripe"
             v-bind:value.sync="valueData"
             v-bind:variant="variant"
             v-bind:border="border"
@@ -12,9 +16,9 @@
             v-bind:disabled="disabled"
             v-bind:width="width"
             v-bind:height="height"
-            v-on:update:value="$emit('update:value', $event)"
-            v-on:blur="$emit('blur')"
-            v-on:focus="$emit('focus')"
+            v-on:update:value="onInput($event)"
+            v-on:blur="onBlur"
+            v-on:focus="onFocus"
         />
         <div class="symbol" v-bind:style="style">
             {{ symbol }}
@@ -29,12 +33,24 @@
     opacity: 0.4;
 }
 
-.input-symbol .input-ripe {
+.input-symbol .input {
     border-radius: 6px 0px 0px 6px;
     padding-right: 7px;
     text-align: right;
     text-overflow: ellipsis;
     vertical-align: top;
+}
+
+.input-symbol.text-align-left .input {
+    text-align: left;
+}
+
+.input-symbol.text-align-center .input {
+    text-align: center;
+}
+
+.input-symbol.text-align-right .input {
+    text-align: right;
 }
 
 .input-symbol .symbol {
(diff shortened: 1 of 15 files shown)


vue/components/ui/organisms/input-symbol/input-symbol.vue
```diff
@@ -0,0 +1,128 @@
...
+            v-bind:value.sync="valueData"
+            v-bind:variant="variant"
+            v-bind:border="border"
+            v-bind:placeholder="placeholder"
+            v-bind:disabled="disabled"
+            v-bind:width="width"
+            v-bind:height="height"
+            v-on:update:value="$emit('update:value', $event)"
+            v-on:blur="$emit('blur')"
+            v-on:focus="$emit('focus')"
```
gcandal: use `onX` methods

vue/components/ui/organisms/input-symbol/input-symbol.vue
```diff
@@ -0,0 +1,128 @@
+<template>
+    <div
+        class="input-symbol"
+        v-bind:class="[{ disabled: disabled }, `color-${variant}`, `border-${border}`]"
+    >
+        <input-ripe
+            class="input-ripe"
```
gcandal: No need, already defined in the input.

vue/components/ui/organisms/input-symbol/input-symbol.vue
```diff
@@ -0,0 +1,128 @@
...
+@import "css/variables.scss";
+
+.input-symbol.disabled {
+    opacity: 0.4;
+}
+
+.input-symbol .input-ripe {
+    border-radius: 6px 0px 0px 6px;
+    padding-right: 7px;
+    text-align: right;
```
gcandal: This should be a prop of the input.

vue/components/ui/organisms/input-symbol/input-symbol.vue
```diff
@@ -0,0 +1,128 @@
...
+    text-overflow: ellipsis;
```
gcandal: Does this do anything?

vue/components/ui/organisms/input-symbol/input-symbol.vue
```diff
@@ -0,0 +1,128 @@
...
+    vertical-align: top;
```
gcandal: Does this do anything?

vue/components/ui/organisms/input-symbol/input-symbol.vue
3rdvision: Yes, when the number is too big, adds "..."
Useful because width is definable prop and it can be too small for the number in question.
As we are working with numbers I thought it was critical to add the `ellipsis`.
imagem

vue/components/ui/organisms/input-symbol/input-symbol.vue
3rdvision: Yes, aligns `input-symbol` in the same horizontal level as `input-ripe`
imagem

vue/components/ui/organisms/input-symbol/input-symbol.vue
gcandal: Move it to the input then :)

3rdvision: Alright :)
I'll do this in a different PR.

gcandal:
@3rdvision take a look at the changes I did to fix the width specification via props.

3rdvision:
> @3rdvision take a look at the changes I did to fix the width specification via props.

I just noticed two issues after those changes:
1- Currency symbols will not show properly if there are more than 2 characters (picture below)*
2- The `width` prop only applies if it's value is higher than 193. This can be a problem specially because the design defined 133px for the whole field which is now not possible to archieve only with props.

* - I implemented input-currency to allow more than 1 character to show as there are currency symbols that are more than 1 charecter long (https://transferwise.com/gb/blog/world-currency-symbols), for instance "kr" for Sweedish krona.

imagem

PR #169, opened 2020-01-09
Add slots to "table-ripe" component

NFSS10:
Added slot that made it possible to replace the table `<tr></tr>`'s


gcandal:
@NFSS10 please add an example of this to the storybook.

gcandal:
Closing in favor of a slightly different approach: https://github.com/ripe-tech/ripe-components-vue/pull/171

PR #173, opened 2020-01-13
Added new slot to the detail header buttons

FlaviooLima:
| - | - |
| --- | --- |
| Issue | There was a need to create an option for adding more buttons in the detail header. |
| Decisions |  Added new slot to the detail header buttons, so we can add in the same div has in the already existing buttons |
| Animated GIF |   |

### DIFF
--- a/vue/components/ui/organisms/details/details.vue
+++ b/vue/components/ui/organisms/details/details.vue
@@ -68,7 +68,9 @@
             <div class="container-header">
                 <div class="header-buttons">
                     <slot name="header-buttons">
-                        <slot name="header-buttons-extra" />
+                        <div class="header-button" v-if="!isMobileWidth()">
+                            <slot name="header-buttons-extra" />
+                        </div>
                         <div class="header-button">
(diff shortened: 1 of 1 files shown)


vue/components/ui/organisms/details/details.vue
```diff
@@ -68,6 +68,7 @@
             <div class="container-header">
                 <div class="header-buttons">
                     <slot name="header-buttons">
+                        <slot name="header-buttons-extra" />
```
joamag: header buttons before
and header buttons after

FlaviooLima: Done!

vue/components/ui/organisms/details/details.vue
```diff
@@ -112,6 +113,7 @@
                             </span>
                             <p>Status</p>
                         </div>
+                        <slot name="header-buttons-after" v-if="!isMobileWidth()" />
```
joamag: move the v-if mobile to the slot user

PR #175, opened 2020-01-13
Table variants & aligment props

BeeMargarida:
| - | - |
| --- | --- |
| Issue | -- |
| Decisions | Added prop variant to table component as well as styling of dense variant value. Also added alignment prop, that aligns the content of the table. |
| Animated GIF | Jan-13-2020 10-49-19|


vue/components/ui/molecules/table/table.vue
```diff
@@ -196,6 +205,11 @@
     width: 20px;
 }
 
+.table[style="text-align: left;"] .table-column > span::before {
```
gcandal: Instead of this, add a class (`table-align-left`) that represents the alignment and use it here

vue/components/ui/molecules/table/table.vue
```diff
@@ -82,16 +82,25 @@
...
+    font-weight: 600;
+}
+
 .table ::v-deep td {
     font-size: 14px;
     font-weight: 600;
     height: 80px;
     overflow: hidden;
-    padding: 0px 0px 0px 0px;
+    padding: 0px 20px 0px 20px;
```
gcandal: Why?

BeeMargarida: When aligned left or right, the text in the body of the table isn't aligned with the table header. This was a way of styling independently of the text-alignment being made.

gcandal:
@BeeMargarida I think you're missing the alignment of the caret:

image

PR #177, opened 2020-01-13
Add ellipsis to input-ripe

3rdvision:
| - | - |
| --- | --- |
| Issue | - |
| Decisions | As suggested in a PR's comment https://github.com/ripe-tech/ripe-components-vue/pull/168#discussion_r365787877, `text-overflow: ellipsis` should be default in `input-ripe`. |
| Animated GIF | imagem |


gcandal:
Guard this behavior behind a prop, which can have `default: true`.

PR #178, opened 2020-01-13
Add listing buttons

gcandal:
Allows to easily add buttons next to the search bar. Example: https://github.com/ripe-tech/ripe-pulse/pull/87


joamag:
Closing this as a new approach is going to be taken

PR #179, opened 2020-01-14
"lineup" component style

NFSS10:
| - | - |
| --- | --- |
| Issue | - |
| Decisions | Improved lineup style and also added prop "nrColumns" that allows to define how many columns should the lineup component have. |
| Animated GIF | **Nr Columns:**<br>lineup_nrColumns**Responsive:**<br>lineup_responsive |

### DIFF
--- a/vue/components/ui/molecules/filter/filter.vue
+++ b/vue/components/ui/molecules/filter/filter.vue
@@ -17,7 +17,7 @@
                 v-bind:items="items"
                 v-bind:values="values"
                 v-bind:get-item-url="getItemUrl"
-                v-bind:nr-columns="lineupNrColumns"
+                v-bind:columns="lineupColumns"
                 v-on:click="onLineupClick"
             >
                 <slot v-bind:name="slot" v-for="slot in Object.keys($slots)" v-bind:slot="slot" />
@@ -111,7 +111,7 @@ export const Filter = {
             type: Boolean,
             default: false
         },
-        lineupNrColumns: {
+        lineupColumns: {
             type: Number,
             default: null
         }
--- a/vue/components/ui/molecules/filter/filter.vue
+++ b/vue/components/ui/molecules/filter/filter.vue
@@ -18,6 +18,7 @@
                 v-bind:values="values"
                 v-bind:get-item-url="getItemUrl"
                 v-on:click="onLineupClick"
+                v-bind:nr-columns="lineupNrColumns"
             >
                 <slot v-bind:name="slot" v-for="slot in Object.keys($slots)" v-bind:slot="slot" />
                 <template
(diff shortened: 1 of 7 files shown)


vue/components/ui/organisms/listing/listing.vue
```diff
@@ -246,6 +247,9 @@ export const Listing = {
         searchWidth: {
             type: Number,
             default: 304
+        },
+        lineupNrColumns: {
+            type: Number
```
gcandal: Either `required: true` or `default: something`. In this case let's go with `default: null`, delegating to the lineup whatever the default should be.

vue/components/ui/molecules/filter/filter.vue
```diff
@@ -109,6 +110,9 @@ export const Filter = {
         useQuery: {
             type: Boolean,
             default: false
+        },
+        lineupNrColumns: {
+            type: Number
```
gcandal: Either `required: true` or `default: something`. In this case let's go with `default: null`, delegating to the lineup whatever the default should be.

vue/components/ui/molecules/lineup/lineup.vue
```diff
@@ -85,56 +86,50 @@
     text-align: left;
 }
 
-.lineup > .lineup-item {
+.lineup .lineup-item {
```
gcandal: Why the change?

vue/components/ui/organisms/listing/listing.vue
NFSS10: Setting it `default: null `will make the prop in lineup be `null`. That's why I didn't added a default to the "lineupNrColumns" prop, as I wanted the default value of "nrColumns" in "lineup" component be controlled by itself.

vue/components/ui/molecules/filter/filter.vue
NFSS10: Same as said before

> Setting it default: null will make the prop in lineup be null. That's why I didn't added a default to the "lineupNrColumns" prop, as I wanted the default value of "nrColumns" in "lineup" component be controlled by itself.

gcandal:
@NFSS10 we'll need you to make the same adaptation you did to RIPE Pulse to other projects (Sputnik, Orchestra, White) too.

vue/components/ui/molecules/lineup/lineup.vue
NFSS10: Redid the style, no particular reason.

gcandal: Revert it then, let's keep following the principle of least action. Unless something is a clear improvement we'd be better to avoid changing it.

gcandal:
@NFSS10 some deleted `>`s are still polluting the diff:
image

PR #182, opened 2020-01-15
Changed Details item prop required to false

FlaviooLima:
| - | - |
| --- | --- |
| Issue | Props from component details required the item to exist from the start. But this creates an error in the majority of cases where we have a parent still requesting the item itself. |
| Decisions |  Remove the required from props so that no error appear. <br> It was  tested in the pulse project -> order show component by forcing the data request to always return null. No bugs where found in the UI or console logs. |

### DIFF
--- a/vue/components/ui/organisms/details/details.vue
+++ b/vue/components/ui/organisms/details/details.vue
@@ -491,7 +491,7 @@ export const Details = {
         },
         item: {
             type: Object,
-            required: true
+            required: false
         },
         index: {
             type: Number,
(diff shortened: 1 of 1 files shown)


gcandal:
@FlaviooLima use `default: null` instead.

Vue's default is `default: undefined` and `required: false`, and what we're doing is using `null` to represent something that is optional and is not defined.

vue/components/ui/organisms/details/details.vue
```diff
@@ -491,7 +491,7 @@ export const Details = {
         },
         item: {
             type: Object,
-            required: true
+            required: false
```
joamag: just remove the required line no need to set it to `false`

PR #166, opened 2020-01-08
#56 Persistent filter

NFSS10:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-pulse/issues/56 |
| Decisions | • Added a select that when pressed opens a dropdown with various filters. The user then can choose a filter. The user can also edit, delete and create new filters. <br>• Depends on the addition of slots in "checkbox" component implemented in PR: https://github.com/ripe-tech/ripe-components-vue/pull/192 |
| Animated GIF | persistent_filters_1<br>**Save new filter modal:**<br>persitent-filters_imp2 <br>**Creating a filter that already exists, updates the existing one:** persistent-filters_imp3<br>**Updates filter based on url:**<br>persistent_filters_url |

PR #184, opened 2020-01-15
Created "search-persistent-local-storage" component

NFSS10:
| - | - |
| --- | --- |
| Issue | - |
| Decisions | • This component works like the "search-persistent" component, but the filters are saved to the user session.<br>• Depends on the implementation of "search-persistent" on PR: https://github.com/ripe-tech/ripe-components-vue/pull/166.|
| Animated GIF | - |


vue/components/ui/molecules/search-persistent-local-storage/search-persistent-local-storage.vue
```diff
@@ -0,0 +1,94 @@
+<template>
+    <div class="search-persistent-local-storage">
+        <global />
```
gcandal: Why is this needed?

vue/components/ui/molecules/search-persistent-local-storage/search-persistent-local-storage.stories.js
```diff
@@ -0,0 +1,100 @@
...
+                    v-bind:value.sync="valueData"
+                    v-bind:filters.sync="filtersData"
+                    v-bind:placeholder="placeholder"
+                    v-bind:icon-visible="iconVisible"
+                    v-bind:clear-visible="clearVisible"
+                    v-bind:variant="variant"
+                    v-bind:width="width"
+                    v-bind:loading="loading"
+                    />
+                <p>Value: {{ valueData }}</p>
```
gcandal: Does this work without emitting anything in the component itself? Do you know why?

vue/components/ui/molecules/search-persistent-local-storage/search-persistent-local-storage.vue
```diff
@@ -0,0 +1,94 @@
...
+        <search-persistent
+            v-bind:value.sync="valueData"
+            v-bind:filters.sync="filtersData"
+            v-bind:placeholder="placeholder"
+            v-bind:icon-visible="iconVisible"
+            v-bind:clear-visible="clearVisible"
+            v-bind:variant="variant"
+            v-bind:width="width"
+            v-bind:loading="loading"
+            v-on:update:filters="onUpdateFilters"
```
gcandal: Either use `sync` or listen to this event. You could use `sync` and then watch on `filtersData`, for instance.

vue/components/ui/molecules/search-persistent-local-storage/search-persistent-local-storage.vue
```diff
@@ -0,0 +1,94 @@
...
+        valueData(value) {
+            this.$emit("update:value", this.valueData);
+        },
+        filtersData(value) {
+            this.$emit("update:filters", this.filtersData);
+        }
+    },
+    computed: {
+        filtersKey() {
+            return this.$root.account ? `filters${this.$root.account._id}` : "filters";
```
gcandal: Is `this.$root.account` reactive? Please use the debugger to find out if it's an `Observable`. If it's not then `filtersKey` will need to be a method instead. Remember the `alert` not getting reset bug? It was caused by a similar problem (in that case, the `$refs` not being reactive).

gcandal:
Nice PR 👍

vue/components/ui/molecules/search-persistent-local-storage/search-persistent-local-storage.vue
NFSS10: I checked, it is reactive.
Atleast in my knowledge, if it has the `__ob__` property it is reactive.

gcandal: Confirmed: https://github.com/ripe-tech/ripe-pulse/blob/2ce3ade6e91c18aa81b623f15a8db0e192f031c8/js/main.js#L207

vue/components/ui/molecules/search-persistent-local-storage/search-persistent-local-storage.vue
gcandal: Ok but that only means that whoever uses this component must include global. If we have a global in every component that defeats the purpose.

Move this to the story instead.

NFSS10: Right, Got it ! 👍

gcandal:
@NFSS10 let's park this until https://github.com/ripe-tech/ripe-components-vue/pull/166 is done. In any case, we'll have to extend this to use a key-by depending on the filter tenancy type, and not only user.

joamag:
Closing with reaper

PR #185, opened 2020-01-15
Change Detail value verification

FlaviooLima:
| - | - |
| --- | --- |
| Issue | Value verification was showing false negatives. so if a value was `0` it would behave like it didn't had a value and it would return the no value default `-` |
| Decisions |  Change the verification to check for only `undefined` and `null` |
| Animated GIF | Jan-15-2020 12-09-39 |


gcandal:
@FlaviooLima can't this be adapted to other components? Please search for `-` and apply the fix where applicable.

FlaviooLima:
@gcandal found two more!

gcandal:
@FlaviooLima the Github build seems to be failing, please take a look (should be a matter of just `npm run pretty`).

PR #189, opened 2020-01-16
Dropdown alignment

NFSS10:
| - | - |
| --- | --- |
| Issue | - |
| Decisions | Allows to specify the select button width, dropdown width and dropdown alignment (left or right). |
| Animated GIF | dropdown_alignment |

### DIFF
--- a/vue/components/ui/atoms/dropdown/dropdown.vue
+++ b/vue/components/ui/atoms/dropdown/dropdown.vue
@@ -177,7 +177,7 @@ export const Dropdown = {
 
             return base;
         },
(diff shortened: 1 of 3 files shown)


gcandal:
@NFSS10 I'm closing this since yesterday I had already done changes to accomplish this in https://github.com/ripe-tech/ripe-components-vue/pull/166 (you're right that it should've been a separate PR though).

The approach is different from yours in this sense: it's the `select` and not the `dropdown` that should have the alignment since it's him who controls the `dropdown` position.

PR #192, opened 2020-01-17
Add slot to "checkbox" component

NFSS10:
| - | - |
| --- | --- |
| Issue | - |
| Decisions |  Added slots to the "checkbox" component. |
| Animated GIF | **Example of a select that was added to the slot, that appears when the choice is checked:**<br>checkbox_slot_example |

### DIFF
--- a/vue/components/ui/atoms/checkbox/checkbox.stories.js
+++ b/vue/components/ui/atoms/checkbox/checkbox.stories.js
@@ -82,7 +82,14 @@ storiesOf("Atoms", module)
                         v-bind:values.sync="valuesData"
                         v-bind:disabled="disabled"
                         v-bind:error="error"
-                    />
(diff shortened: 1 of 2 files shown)


gcandal:
@NFSS10 for now this seems a bit too specific to be in the component. Do you see this kind of design pattern being used elsewhere? For now this should be just in the persistent filter modal.

NFSS10:
> @NFSS10 for now this seems a bit too specific to be in the component. Do you see this kind of design pattern being used elsewhere? For now this should be just in the persistent filter modal.

I did the same as I did when adding the chat to ripe-pulse. Making it this way, supports adding anything for every checkbox.

joamag:
@gcandal @NFSS10 need to port the exact same solution for the radio buttons

PR #193, opened 2020-01-20
Fix input width

3rdvision:
| - | - |
| --- | --- |
| Issue | When we pass a width prop lower than 203 to `input-symbol`, it doesn't size as it should. This is because input has a default browser's `width` and flexbox's `min-width: auto`. This can be solved by overriding the behavior  with `min-width: 0px`.  Further details at: https://stackoverflow.com/a/42421490   |
| Decisions |   |
| Animated GIF | Below  |

### Before
`width = 100`
imagem

### After
`width = 100`
imagem

### DIFF
--- a/vue/components/ui/atoms/input/input.vue
+++ b/vue/components/ui/atoms/input/input.vue
@@ -28,6 +28,7 @@
     height: 34px;
     letter-spacing: 0.25px;
     line-height: 34px;
+    min-width: 0px;
     outline: none;
     padding-left: 12px;
     padding-right: 12px;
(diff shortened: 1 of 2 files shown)


vue/components/ui/atoms/input/input.vue
```diff
@@ -28,6 +28,7 @@
     height: 34px;
     letter-spacing: 0.25px;
     line-height: 34px;
+    min-width: 0px;
```
joamag: @gcandal why this?

PR #194, opened 2020-01-20
Persitent Filters - Tenancy improvements

NFSS10:
| - | - |
| --- | --- |
| Issue | Related to issue https://github.com/ripe-tech/ripe-components-vue/issues/166 |
| Decisions |  • Depends on the addition of slots in "checkbox" component implemented in PR: https://github.com/ripe-tech/ripe-components-vue/pull/192 <br>• Made the improvements specified in the following comments: [Comment 1](https://github.com/ripe-tech/ripe-components-vue/pull/166#issuecomment-575211514), [Comment 2](https://github.com/ripe-tech/ripe-components-vue/pull/166#issuecomment-575545847) |
| Animated GIF | persitent-filters_imp2 <br>**Creating a filter that already exists, updates the existing one:** persistent-filters_imp3|

### DIFF
--- a/vue/components/ui/atoms/dropdown/dropdown.vue
+++ b/vue/components/ui/atoms/dropdown/dropdown.vue
@@ -224,7 +224,7 @@ export const Dropdown = {
             this.dehighlight(index);
         },
         onSlideAfterLeave() {
-            this.$emit("animation:ended");
+            this.$emit("animation:close:ended");
         },
         _getItemClasses(item, index) {
             return {
--- a/vue/components/ui/atoms/dropdown/dropdown.vue
+++ b/vue/components/ui/atoms/dropdown/dropdown.vue
@@ -1,7 +1,7 @@
 <template>
     <div class="dropdown-container">
         <global-events v-on:keydown.esc="handleGlobal()" />
-        <transition name="slide">
+        <transition name="slide" v-on:after-leave="onSlideAfterLeave">
             <ul class="dropdown" v-bind:style="dropdownStyle" v-show="visibleData" ref="dropdown">
                 <li
                     class="dropdown-item"
@@ -222,6 +222,9 @@ export const Dropdown = {
         },
         onMouseleave(index) {
             this.dehighlight(index);
+        },
+        onSlideAfterLeave() {
+            this.$emit("animation:ended");
         },
         _getItemClasses(item, index) {
             return {
--- a/vue/components/ui/molecules/search-persistent/save-filter-modal.vue
+++ b/vue/components/ui/molecules/search-persistent/save-filter-modal.vue
@@ -107,7 +107,7 @@ export const SaveFilterModal = {
             visibleSelects: new Array(this.tenancyItems.length).fill(false),
             tenacyValuesData: {},
             filterNameData: null,
-            searchData: this.search,
+            searchData: this.search
         };
     },
     watch: {
@@ -206,7 +206,7 @@ export const SaveFilterModal = {
         onSaveClick() {
             this.$emit("click:confirm");
         },
-        onUpdateSelectVisible(value, index){
+        onUpdateSelectVisible(value, index) {
             this.closeAllSelects();
             this.setSelectVisible(index, value);
         },
--- a/vue/components/ui/molecules/search-persistent/save-filter-modal.vue
+++ b/vue/components/ui/molecules/search-persistent/save-filter-modal.vue
@@ -24,13 +24,14 @@
             <checkbox v-bind:items="tenancyItems" v-bind:values.sync="tenacyValuesData">
                 <template v-slot:extra-info="{ item, index }">
                     <select-ripe
-                        v-if="getTenancySelectOptions(item.label)"
+                        v-if="getTenancySelectOptions(item.value)"
+                        v-show="isTenancySelected(item.value)"
                         v-bind:placeholder="selectPlaceholder(item)"
                         v-bind:width="200"
                         v-bind:align="'left'"
                         v-bind:max-height="150"
-                        v-bind:options="getTenancySelectOptions(item.label)"
-                        v-on:update:value="(value, selectedItemIndex) => onSelected(item.label, value, selectedItemIndex)"
+                        v-bind:options="getTenancySelectOptions(item.value)"
+                        v-on:update:value="(value, selectedItemIndex) => onSelected(item.value, value, selectedItemIndex)"
                     />
                 </template>
             </checkbox>
@@ -157,16 +158,19 @@ export const SaveFilterModal = {
                 { value: "factory_e", label: "Factory E" }
             ];
         },
-        getTenancySelectOptions(tenancy) {
-            switch (tenancy) {
-                case "Brand":
+        getTenancySelectOptions(tenancyValue) {
+            switch (tenancyValue) {
+                case "brand":
                     return this.getBrands();
-                case "Channel":
+                case "channel":
                     return this.getChannels();
-                case "Factory":
+                case "factory":
                     return this.getFactories();
                 default: return null;
             }
+        },
+        isTenancySelected(tenancyValue) {
+            return Boolean(this.tenacyValuesData[tenancyValue]);
         },
         selectPlaceholder(item) {
             return `Select ${item.label}`;
@@ -177,9 +181,9 @@ export const SaveFilterModal = {
         onSaveClick() {
             this.$emit("click:confirm");
         },
-        onSelected(tenancy, value, selectedValueIndex) {
+        onSelected(tenancyValue, value, selectedValueIndex) {
             //TODO
-            console.log(`Select ${tenancy}: `, selectedValueIndex, value);
+            console.log(`Select ${tenancyValue}: `, selectedValueIndex, value);
         }
     }
 };
--- a/vue/components/ui/molecules/search-persistent/save-filter-modal.vue
+++ b/vue/components/ui/molecules/search-persistent/save-filter-modal.vue
@@ -23,7 +23,7 @@
         <form-input v-bind:header="'Tenancy'" v-if="checkableTenancyChoices.length !== 1">
             <checkbox
                 v-bind:items="checkableTenancyChoices"
-                v-bind:values.sync="tenancyCheckboxValuesData"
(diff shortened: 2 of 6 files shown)


vue/components/ui/atoms/dropdown/dropdown.vue
```diff
@@ -223,6 +223,9 @@ export const Dropdown = {
         onMouseleave(index) {
             this.dehighlight(index);
         },
+        onSlideAfterLeave() {
+            this.$emit("animation:ended");
```
gcandal: This only happens when closing, right? If that's the case, suffix it with `close` .

vue/components/ui/molecules/select/select.vue
```diff
@@ -158,6 +159,10 @@ export const Select = {
             type: Boolean,
             default: false
         },
+        width: {
```
gcandal: Dead code?

vue/components/ui/molecules/search-persistent/search-persistent.vue
```diff
@@ -272,13 +274,28 @@ export const SearchPersistent = {
...
+        },
         addFilters(filters) {
+            for (let i = 0; i < filters.length; i++) {
+                const index = this.filterExists(filters[i]);
+
+                if (index >= 0) {
+                    this.updateFilter(filters[i], index);
+                    filters.splice(i, 1);
+                }
+            }
```
gcandal: What is this used for?

vue/components/ui/molecules/search-persistent/save-filter-modal.vue
```diff
@@ -95,25 +121,148 @@ export const SaveFilterModal = {
...
-    mounted: async function() {},
+    mounted: async function() {
+        this.tenancies = this.builtTenancies;
+
+        // Getting brands, channels and factories
+        this.tenancies.brand.choices = await this.getBrands();
+        this.tenancies.channel.choices = await this.getChannels();
+        this.tenancies.factory.choices = await this.getFactories();
+
+        this.removeInvalidChoices();
```
gcandal: Use a computed instead to achieve the same result.

vue/components/ui/atoms/dropdown/dropdown.vue
NFSS10: Yes, it triggers after the close animation.

vue/components/ui/molecules/search-persistent/save-filter-modal.vue
```diff
@@ -95,25 +121,148 @@ export const SaveFilterModal = {
...
+        isTenancyChoiceValid(value) {
+            return value === "user" || this.hasTenancyItems(value);
+        },
+        isTenancySelectValid(value) {
+            return this.hasTenancyItems(value) && this.hasRequiredItemsLength(value);
+        },
+        hasTenancyItems(value) {
+            return Boolean(this.tenancies[value] && this.tenancies[value].choices.length);
+        },
+        hasRequiredItemsLength(value) {
```
gcandal: Suggested renaming to avoid having to read the implementation to know what the functions do:
* `hasChoice(tenancy)`
* `hasChoices(tenancy)`

The names might not exactly be these, but something that makes it clear that the functions are almost the same.

vue/components/ui/molecules/search-persistent/save-filter-modal.vue
```diff
@@ -95,25 +121,148 @@ export const SaveFilterModal = {
...
+                }
+            }
+        },
+        isTenancyChoiceSelected(value) {
+            return Boolean(this.tenancyCheckboxValuesData[value]);
+        },
+        isTenancyChoiceValid(value) {
+            return value === "user" || this.hasTenancyItems(value);
+        },
+        isTenancySelectValid(value) {
```
gcandal: You could just use `hasRequiredItemsLength` right?

vue/components/ui/molecules/search-persistent/search-persistent.vue
NFSS10: To handle this: "Create filter with duplicated name = update existing filter"

This code checks for each newly created filter if it already exists. If yes updates it and removes him from the filters array so that they arent added again.

vue/components/ui/molecules/search-persistent/save-filter-modal.vue
```diff
@@ -95,25 +121,148 @@ export const SaveFilterModal = {
...
+                    this.tenancyItemsData.splice(i, 1);
+                    i--;
+                }
+            }
+        },
+        isTenancyChoiceSelected(value) {
+            return Boolean(this.tenancyCheckboxValuesData[value]);
+        },
+        isTenancyChoiceValid(value) {
+            return value === "user" || this.hasTenancyItems(value);
```
gcandal: Inline this code. I appreciate trying to use small functions, but when the intent is not clear from the name then I'd rather have it inline.

vue/components/ui/molecules/search-persistent/save-filter-modal.vue
```diff
@@ -95,25 +121,148 @@ export const SaveFilterModal = {
...
+            // TODO fix me, I'm hardcoded and have 5 option
+            return [
+                { value: "factory_a", label: "Factory A" },
+                { value: "factory_b", label: "Factory B" },
+                { value: "factory_c", label: "Factory C" },
+                { value: "factory_d", label: "Factory D" },
+                { value: "factory_e", label: "Factory E" }
+            ];
+        },
+        removeInvalidChoices() {
```
gcandal: Can you find a better name? They're not actually invalid, they just have <=1 options.

vue/components/ui/molecules/search-persistent/save-filter-modal.vue
```diff
@@ -95,25 +121,148 @@ export const SaveFilterModal = {
...
             );
         },
         filters() {
-            return Object.keys(this.tenacyValuesData).map(tenancy => ({
+            return Object.keys(this.tenancyCheckboxValuesData).map(tenancy => ({
                 name: this.filterNameData,
                 value: this.searchData,
-                tenancy: tenancy,
+                tenancy: this.getSelectedTenancy(tenancy),
                 context: "" // TODO
```
gcandal: The tenancy is factory/channel/brand, it's the context that must have the corresponding value.

vue/components/ui/molecules/search-persistent/save-filter-modal.vue
```diff
@@ -95,25 +121,148 @@ export const SaveFilterModal = {
...
             }));
+        },
+        isTenancyEmpty() {
+            return this.tenancyItemsData.length === 1;
+        },
```
gcandal: Inline this function. I understand what you mean but saying "empty" in the method and then having `=== 1` is a bit odd, but if you have it in the proper context (HTML) then it becomes clear what you're doing.

vue/components/ui/molecules/search-persistent/save-filter-modal.vue
```diff
@@ -95,25 +121,148 @@ export const SaveFilterModal = {
...
         }
     },
-    mounted: async function() {},
+    mounted: async function() {
+        this.tenancies = this.builtTenancies;
+
+        // Getting brands, channels and factories
+        this.tenancies.brand.choices = await this.getBrands();
+        this.tenancies.channel.choices = await this.getChannels();
+        this.tenancies.factory.choices = await this.getFactories();
```
gcandal: I understand you're pushing for code reuse with this approach, but I got the extreme feeling that the solution got over-engineered because of this. On the one hand you're trying to make this general for N tenancy types, but that premise is broken since you have to hard-code the `getX` anyway.

Split this into `brandChoices`, `channelChoices`, etc. Then every time you need to make common assertions, you just need to pass `brandChoices` instead of passing `"brand"` and then acessing `tenancies`.

Trust me that once you make this seemingly small change the overall structure will get much simpler.

vue/components/ui/molecules/search-persistent/save-filter-modal.vue
```diff
@@ -95,25 +121,148 @@ export const SaveFilterModal = {
...
+        userTenancy() {
+            return { choices: [], selectedValue: "user", selectVisible: false };
+        },
+        builtTenancies() {
+            const builtTenancies = {};
+
+            this.tenancyItems.forEach(item => {
+                if (item.value === "user") builtTenancies[item.value] = this.userTenancy;
+                else {
+                    builtTenancies[item.value] = {
```
gcandal: In the same spirit as https://github.com/ripe-tech/ripe-components-vue/pull/194/files#r368594708 , try to solve this without these deeply nested data structures.

vue/components/ui/molecules/search-persistent/save-filter-modal.vue
NFSS10: This was the code before I started simplifying things: https://github.com/ripe-tech/ripe-components-vue/commit/02789d3c04a30b93ab02c2bfcba1d75a1fdb1bf7 and https://github.com/ripe-tech/ripe-components-vue/commit/6dd56baa76942bfa183a7b5a0258ba8c74118b90

I changed it because I thought it was a little confusing to read the code and I felt like I had a lot of similar logic, so I tried to make simplify this and make it more reusable.

I made the changes so that I could just use a computed like the `builtTenancies`to create the data structures needed. After that I would just need to worry about how to load the choices and the code would work.

vue/components/ui/molecules/select/select.vue
NFSS10: This prop was probably removed as there is code that uses it. See https://github.com/ripe-tech/ripe-components-vue/blob/a1ce5d7296d0d134c71c5ad3104bd5c249b439e5/vue/components/ui/molecules/select/select.vue#L383
Also the prop width is being used here for example: https://github.com/ripe-tech/ripe-components-vue/blob/a1ce5d7296d0d134c71c5ad3104bd5c249b439e5/vue/components/ui/molecules/search-persistent/save-filter-modal.vue#L32

vue/components/ui/molecules/search-persistent/save-filter-modal.vue
NFSS10: Yes, my bad.

PR #198, opened 2020-01-21
Fix table item click index

3rdvision:
| - | - |
| --- | --- |
| Issue | When a table row is clicked, `table-ripe` emits a click event `$emit("click", item, index)`, and is expected that the payload of that event is the clicked item and the item's index position in the `items` passed as prop. Instead, it is emiting the position index in the table. This is a problem when the table has different a orientation than the original when sorted and/or reversed. With this fix `$emit("click", item, index)` will emit in the click event, the clicked item and the respective item's index. |
| Decisions |   |
| Animated GIF |   |

### DIFF
--- a/vue/components/ui/molecules/table/table.vue
+++ b/vue/components/ui/molecules/table/table.vue
@@ -260,12 +260,15 @@ export const Table = {
         };
     },
     computed: {
+        itemsWithIndex() {
+            return this.items.map((item, index) => ({ _originalIndex: index, ...item }));
+        },
         sortedItems() {
             if (!this.sortData) {
                 return this.items;
             }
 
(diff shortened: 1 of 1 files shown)


vue/components/ui/molecules/table/table.vue
```diff
@@ -280,8 +283,8 @@ export const Table = {
             this.$emit("update:sort", this.sortData);
             this.$emit("update:reverse", this.reverseData);
         },
-        onClick(item, index) {
-            this.$emit("click", item, index);
+        onClick(item) {
+            this.$emit("click", item, item._originalIndex);
```
gcandal: Include the current index too, it might be useful.

PR #200, opened 2020-01-21
Add to table mouseover and header height prop

FlaviooLima:
| - | - |
| --- | --- |
| Issue | We need `mouseover` and a prop to set the height of the table header to be able to build the dynamic component |
| Decisions |  Pass those trough props and apply in the code |
| Animated GIF |   |

### DIFF
--- a/vue/components/ui/molecules/table/table.vue
+++ b/vue/components/ui/molecules/table/table.vue
@@ -26,7 +26,7 @@
                 <tr
                     v-bind:key="item.id"
                     v-on:click="onClick(item, index)"
-                    v-on:mouseover="mouseOver(item, index)"
+                    v-on:mouseover="onMouseOver(item, index)"
                 >
                     <slot v-bind:item="item" v-bind:index="index">
                         <td
@@ -251,7 +251,7 @@ export const Table = {
         },
         headerHeight: {
             type: Number,
-            default: 36
+            default: null
         }
     },
     watch: {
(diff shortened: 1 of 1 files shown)


vue/components/ui/molecules/table/table.vue
```diff
@@ -244,6 +248,10 @@ export const Table = {
         reverse: {
             type: Boolean,
             default: false
+        },
+        headerHeight: {
+            type: Number,
+            default: 36
```
NFSS10: make it optional: `default: null`.

vue/components/ui/molecules/table/table.vue
```diff
@@ -275,6 +283,9 @@ export const Table = {
             const order = this.reverseData ? "ascending" : "descending";
             return this.sortData === column ? `active ${order}` : "";
         },
+        columnStyle(column) {
+            return { width: column.width, height: `${this.headerHeight}px` };
```
NFSS10: We are using px for sizes. Also, make it optional: 
`            const base = {};
            if (this.width) base.width = `${this.width}px`;
            if (this.headerHeight) base.height = `${this.headerHeight}px`;
            return base;`

vue/components/ui/molecules/table/table.vue
```diff
@@ -283,6 +294,9 @@ export const Table = {
         },
         onClick(item, index) {
             this.$emit("click", item, index);
+        },
+        mouseOver(item, index) {
```
NFSS10: prefix events with on: mouseOver -> onMouseOver

FlaviooLima:
No need for this any more

PR #208, opened 2020-01-23
Form-input inline variant

FlaviooLima:
| - | - |
| --- | --- |
| Issue | We need a input that is inline with the label |
| Decisions |   Added new `Mode` prop to `form-input` that allows to transform the ui of the component |
| Animated GIF |  Jan-23-2020 11-58-06 |


vue/components/ui/molecules/form-input/form-input.vue
```diff
@@ -50,6 +69,15 @@
     line-height: 0px;
 }
 
+.form-input.form-input-inline .content {
+    flex: 1;
```
3rdvision: Add a min-width here as well such that both label and input can shrink at the same time

vue/components/ui/molecules/form-input/form-input.vue
```diff
@@ -1,28 +1,43 @@
 <template>
-    <div class="form-input">
-        <label-ripe class="header" v-bind:text="header" v-bind:for="id" v-if="header" />
-        <div class="content">
-            <slot />
-        </div>
+    <div class="form-input" v-bind:class="alignment">
```
gcandal: Move `alignment` to a `classes` `computed`, similarly to what we have in other components.

vue/components/ui/molecules/form-input/form-input.vue
```diff
@@ -1,28 +1,43 @@
...
-                <label-ripe v-bind:text="error" v-bind:color="'error'" />
-            </div>
-            <div class="text" v-if="warning">
-                <span class="icon icon-warning" />
-                <label-ripe v-bind:text="warning" />
-            </div>
-            <div class="text" v-if="success">
-                <span class="icon icon-success" />
-                <label-ripe v-bind:text="success" v-bind:color="'success'" />
+
```
gcandal: Did you `npm run pretty`? There shouldn't be blank lines.

vue/components/ui/molecules/form-input/form-input.vue
```diff
@@ -99,6 +133,19 @@ export const FormInput = {
         success: {
             type: String,
             default: null
+        },
+        alignment: {
```
gcandal: I understand why you chose this name, but since we're using this for left/center/right in other components let's change it to `variant`, which is the name we're using for "style variations" kind of stuff.

vue/components/ui/molecules/form-input/form-input.vue
```diff
@@ -50,6 +75,15 @@
     line-height: 0px;
 }
 
+.form-input.inline .content {
```
gcandal: `.form-input.form-input-inline`. We usually use this pattern for this kind of stuff (e.g.: `.button-color.button-color-black`).

vue/components/ui/molecules/form-input/form-input.vue
FlaviooLima: Yeah a did, but I doesn't remove the line :/

joamag:
@gcandal we should drop this solution in facvor of improving form input. The input should remain a molecule with much less funcionality

PR #212, opened 2020-01-23
Currency update

3rdvision:
| - | - |
| --- | --- |
| Issue | - |
| Decisions |  Adds more currency symbols according to ISO standard currency codes (https://apps.cybersource.com/library/documentation/sbc/quickref/currencies.pdf); Fix value type to only accept numbers; Fix zero places - when a currency has zero places `return currencyInfo.places || fallback;` would always return `fallback` instead of `currencyInfo.places` holding the value 0. |
| Animated GIF |   |

### DIFF
--- a/vue/components/ui/molecules/input-currency/input-currency.stories.js
+++ b/vue/components/ui/molecules/input-currency/input-currency.stories.js
@@ -16,7 +16,19 @@ storiesOf("Molecules", module)
                     {
                         USD: "USD",
                         EUR: "EUR",
-                        GBP: "GBP"
+                        GBP: "GBP",
+                        CRC: "CRC",
+                        ILS: "ILS",
+                        INR: "INR",
+                        JPY: "JPY",
+                        KRW: "KRW",
+                        NGN: "NGN",
+                        PHP: "PHP",
+                        PLN: "PLN",
+                        PYG: "PYG",
+                        THB: "THB",
+                        UAH: "UAH",
+                        VND: "VND"
                     },
                     "USD"
                 )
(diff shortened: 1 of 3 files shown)


vue/components/ui/molecules/input-currency/input-currency.vue
```diff
@@ -109,7 +109,7 @@ export const InputCurrency = {
         },
         _places(currency, fallback = 2) {
             const currencyInfo = currencies[currency] || {};
-            return currencyInfo.places || fallback;
+            return "places" in currencyInfo ? currencyInfo.places : fallback;
```
gcandal: Your approach is sound, but please change to be similar with the way we usually do this kind of stuff: `currencyInfo.places === undefined ? fallback : currencyInfo.places`

vue/components/ui/molecules/input-currency/input-currency.vue
```diff
@@ -28,7 +28,7 @@ export const InputCurrency = {
             default: () => ({})
         },
         value: {
-            type: String,
+            type: Number,
```
gcandal: Maybe we should accept both `String` and `Number`. Then, we can convert it to a `Number` every time we copy this to `valueData`.

vue/components/ui/molecules/input-currency/input-currency.stories.js
```diff
@@ -8,15 +8,27 @@ storiesOf("Molecules", module)
     .add("Input Currency", () => ({
         props: {
             value: {
-                default: text("Value", "50")
+                default: text("Value", 50)
```
gcandal: `number`

vue/components/ui/molecules/input-currency/input-currency.vue
```diff
@@ -28,7 +28,7 @@ export const InputCurrency = {
             default: () => ({})
         },
         value: {
-            type: String,
+            type: String | Number,
```
joamag: why the or?

gcandal: To allow both types, similarly to what we already do in `input`

PR #233, opened 2020-02-03
Forward input keyup

BeeMargarida:
| - | - |
| --- | --- |
| Issue | -- |
| Decisions | On enter event emitted when enter is clicked in the input. Useful for certain situations, like PR [#24](https://github.com/ripe-tech/ripe-copper/pull/24)  |
| Animated GIF | -- |

### DIFF
--- a/vue/components/ui/atoms/input/input.vue
+++ b/vue/components/ui/atoms/input/input.vue
@@ -11,7 +11,7 @@
         v-on:input="onInput($event.target.value)"
         v-on:focus="onFocus"
         v-on:blur="onBlur"
-        v-on:keyup.enter="onEnter"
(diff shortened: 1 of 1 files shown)


vue/components/ui/atoms/input/input.vue
```diff
@@ -11,6 +11,7 @@
         v-on:input="onInput($event.target.value)"
         v-on:focus="onFocus"
         v-on:blur="onBlur"
+        v-on:keyup.enter="onEnter"
```
gcandal: Let's forward the whole `keyup` instead, it's a little more flexible: `v-on:keyup=stuff / stuff(event) { $emit('keyup', event)`

PR #237, opened 2020-02-06
Emit event in watch of "checkedData" and click event

NFSS10:
| - | - |
| --- | --- |
| Issue | - |
| Decisions | • Emit event in watch of "checkedData"<br>• Emit click event when checkbox is clicked |
| Animated GIF | - |

### DIFF
--- a/vue/components/ui/atoms/checkbox/checkbox.vue
+++ b/vue/components/ui/atoms/checkbox/checkbox.vue
@@ -3,7 +3,7 @@
         tabindex="0"
         class="checkbox"
         v-bind:class="classes"
-        v-on:click="onClick()"
+        v-on:click="onClick"
         v-on:mousedown="onMouseDown()"
         v-on:mouseup="onMouseUp()"
         v-on:keydown.space="onSpace()"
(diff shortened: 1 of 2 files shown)


vue/components/ui/atoms/checkbox/checkbox.vue
```diff
@@ -178,7 +179,6 @@ export const Checkbox = {
             if (this.disabled) return;
 
             this.checkedData = !this.checkedData;
-            this.$emit("update:checked", this.checkedData);
```
gcandal: 👍

gcandal:
@NFSS10 we should stick with a boolean `checked` and just alternate the `icon` for representing "partial" or whatever it is. Any reason for not doing so?

NFSS10:
> @NFSS10 we should stick with a boolean `checked` and just alternate the `icon` for representing "partial" or whatever it is. Any reason for not doing so?

Yes, as stated in the issue description. As it was before this change, to set the checkbox like this  imagem we need to set the props `icon = "minus"` and `checked = true`. This works well on its own but has a side effect when using the checkbox as a global checkbox (see issue description) because when setting `checked = true`, it should select all of the other checkboxes, so there is a conflict here.

gcandal:
@NFSS10 let's talk more about this, please remind me tomorrow

PR #241, opened 2020-02-07
Fix select list syncing

gcandal:


gcandal:
@Benedito0 nicely caught 👍

PR #209, opened 2020-01-23
Checkboxes in "table-ripe" component

NFSS10:
| - | - |
| --- | --- |
| Issue | - |
| Decisions | • Added checkboxes to the "table-ripe" component. Sorting the items doesn't affect the selection <br>• Ctrl + A = Selects all rows<br>• Ctrl + Alt + A = Deselects all rows <br>• Shift + Click and Shift + Click in another checkbox = Selects all the rows checkboxes between the two clicked rows |
| Animated GIF | table_checkboxes <br><hr><br> table-ripe_checkboxes_shift_click|

PR #253, opened 2020-02-14
Selected item is highlighted

NFSS10:
| - | - |
| --- | --- |
| Issue | Seperated this from https://github.com/ripe-tech/ripe-components-vue/pull/209 |
| Decisions | Allows to select a row by its `Id`, giving it an highlight |
| Animated GIF | table-ripe_selected_row |


gcandal:
Closing for now, let's go with having it in the client of the table. Later we'll revisit this if it makes sense.

PR #254, opened 2020-02-17
Add focusable classes

gcandal:


joamag:
@gcandal let's just add the methods and not the classes

vue/components/ui/atoms/checkbox/checkbox.vue
```diff
@@ -3,6 +3,7 @@
         tabindex="0"
         class="checkbox"
         v-bind:class="classes"
+        ref="root"
```
joamag: we're using root as the name?

PR #276, opened 2020-02-24
Smoother progress bar

NFSS10:
| - | - |
| --- | --- |
| Issue | - |
| Decisions | Progress bar is a lot smoother now |
| Animated GIF | **Before**progress-bar_fill_before <br>**After this change**progress-bar_fill_after |


gcandal:
Make it configurable via prop, let it default to `ease` and add it to the storybook in a way that I can see it changing (by using a function to change the progress in a cyclic manner, for instance).

PR #292, opened 2020-04-30
Update table-ripe component

Benedito0:
| - | - |
| --- | --- |
| Issue | |
| Dependencies | |
| Decisions |Copy the template and the logic of ripe-white's [_table-ripe_](https://github.com/ripe-tech/ripe-white/blob/master/vue/components/molecules/table/table.vue) |
| Animated GIF | |

### DIFF
--- a/vue/components/ui/molecules/table/table.vue
+++ b/vue/components/ui/molecules/table/table.vue
@@ -353,7 +353,7 @@ export const Table = {
         },
         classes() {
             const base = {
-                alignment: this.alignment === "right" ? "text-align-right" : ""
+                alignment: this.alignment === "left" ? "text-align-left" : ""
             };
             if (this.variant) base[`table-${this.variant}`] = true;
             return base;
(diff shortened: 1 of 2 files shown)


vue/components/ui/molecules/table/table.vue
```diff
@@ -340,7 +353,7 @@ export const Table = {
         },
         classes() {
             const base = {
-                alignment: this.alignment === "left" ? "text-align-left" : ""
+                alignment: this.alignment === "right" ? "text-align-right" : ""
```
gcandal, line 356: @Benedito0 please revert this part. Even though it's in the "javascript" part, it's actually related with the UI and with the fact that the `table` here has a different default alignment that the table in RIPE White.

gcandal: @Benedito0 this is still unresolved...

Benedito0: Sorry, foi sem querer o request

PR #316, opened 2020-06-02
Port table-expandable from ripe-white

gcandal:
| - | - |
| --- | --- |
| Decisions | Direct port from ripe-white. |


joamag:
@gcandal we need this?

PR #252, opened 2020-02-13
Add filter to select component

3rdvision:
| - | - |
| --- | --- |
| Issue | Add filter mode to `select-ripe`. Also fixes scroll bug. |
| Decisions | - |
| Animated GIF | imagem |

PR #325, opened 2020-06-18
Add select unit tests

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-components-vue/pull/252 |
| Decisions | Add selector tests as suggested in https://github.com/ripe-tech/ripe-components-vue/pull/252#issuecomment-644160963 |

### DIFF
--- a/vue/test/components/select/select.test.js
+++ b/vue/test/components/select/select.test.js
@@ -2,19 +2,19 @@ const assert = require("assert");
 const base = require("../../base");
 
 describe("Select", () => {
-    it("Starts with null value", () => {
+    it("should start with null value", () => {
         const component = base.getComponent("Select");
         const componentData = component.vm.$data;
         assert.strictEqual(componentData.valueData, null);
     });
 
-    it("Starts with no dropdown items", () => {
+    it("should start with no dropdown items", () => {
         const component = base.getComponent("Select");
         const dropdownItemsLength = component.findAll(".dropdown-item").length;
         assert.strictEqual(dropdownItemsLength, 0);
     });
 
-    it("Displays all options labels according to prop", async () => {
+    it("should display all options labels according to prop", async () => {
         const component = base.getComponent("Select", {
             props: {
                 options: [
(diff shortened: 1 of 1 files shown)


joamag:
This seems pretty good 💪

vue/test/components/select/select.test.js
```diff
@@ -0,0 +1,324 @@
...
+                    { value: "option_2", label: "B" },
+                    { value: "option_3", label: "C" }
+                ],
+                value: "option_1"
+            }
+        });
+        const componentData = component.vm.$data;
+        const selectButton = component.get(".select-button");
+        assert.strictEqual(selectButton.text(), "A");
+        assert.strictEqual(componentData.valueData, "option_1");
```
gcandal, line 53: We shouldn't do it like this, because it's not equivalent to what we do when using this component. Tests must rely on public API (props, events, visual state, etc) rather on internal data.

Because of this, this must be changed to bind to the `update:value` event and make assertions based on that. On "real world" usage of this component that's what we do when using `sync`, we never look for internal data.

vue/test/components/select/select.test.js
```diff
@@ -0,0 +1,324 @@
...
+    it("should update visibility on select button click", async () => {
+        const component = base.getComponent("Select");
+        const componentData = component.vm.$data;
+
+        const selectButton = component.get(".select-button");
+        const dropdown = component.findComponent({ name: "dropdown" });
+        const dropdownData = dropdown.vm.$data;
+
+        // toggles visibility on select-button click
+        assert.strictEqual(componentData.visibleData, false);
```
gcandal, line 97: Same as above: we need to really test the visibility of the element, and not it's `visible` data. Check if https://vue-test-utils.vuejs.org/api/wrapper/#isvisible helps.

vue/test/components/select/select.test.js
```diff
@@ -0,0 +1,324 @@
...
+        assert.strictEqual(dropdownData.visibleData, false);
+        await selectButton.trigger("click");
```
gcandal, line 99: Space between actions and assertions, like you did in the rest of the places, so that the end result is something like:

```
action1
assertion about action1
assertion about action1
assertion about action1

action2
assertion about action2
```

Issue #322, opened 2020-06-09
Add label on empty dropdown


gcandal:
##  Description

When a `dropdown` is empty, show a `messageEmpty` message to the user, if it's defined (this new prop is optional).

This should be optional and if it does not exist and the dropdown is empty the dropdown contents should not be visible.

gcandal:
@Benedito0 please port this to RIPE White

PR #329, opened 2020-06-22
#322: Add label on empty dropdown

Benedito0:
| - | - |
| --- | --- |
| Issue | #322  |
| Decisions | Decided to use a li element and check first if the there is no items do the minumum changes in the DOM. Wrote 7 unit tests: ✓ should show a message if there are no items defined. - ✓ should have items accordanly with props.  - ✓ should close on item click. -  ✓ should set item as selected based on the props. - ✓ should highlight item on item mouseover. - ✓ should send the correct highlight event - ✓ should send the correct click event|
| Animated GIF | 6cwtbJHVXl|


joamag:
@gcandal check the unit tests and assign it to me

vue/test/components/dropdown/dropdown.test.js
```diff
@@ -0,0 +1,142 @@
...
+        const component = base.getComponent("Dropdown", {
+            props: { items: items }
+        });
+
+        const numItems = items.length;
+        const itemIndex = Math.floor(Math.random() * numItems);
+        const item = component.findAll(".dropdown-item").at(itemIndex);
+        await item.trigger("mouseenter");
+        await item.trigger("mouseover");
+        assert.strictEqual(component.vm.$data.highlightedData[itemIndex], true);
```
NFSS10, line 95: I think its better to use constant indexes

Also add cases like moving the mouse the another item and seeing if the highlight in both changed and removing the mouse completly from over the items.

vue/test/components/dropdown/dropdown.test.js
```diff
@@ -0,0 +1,142 @@
...
+                value: "text_2"
+            }
+        ];
+        const component = base.getComponent("Dropdown", {
+            props: { items: items }
+        });
+
+        const item = component.findAll(".dropdown-item").at(0);
+        await item.trigger("click");
+        assert.strictEqual(component.vm.$data.visibleData, false);
```
NFSS10, line 57: Check if if the component is openend to ensure the click is the one that changed `visibleData` state

vue/test/components/dropdown/dropdown.test.js
```diff
@@ -0,0 +1,142 @@
...
+            {
+                value: "text_2"
+            }
+        ];
+        const component = base.getComponent("Dropdown", {
+            props: { items: items }
+        });
+
+        const numItems = items.length;
+        const itemIndex = Math.floor(Math.random() * numItems);
```
NFSS10, line 113: I think its better to use constant indexes to ensure the test will run the same way everytime

vue/test/components/dropdown/dropdown.test.js
```diff
@@ -0,0 +1,142 @@
...
+        const item = component.findAll(".dropdown-item").at(itemIndex);
+        await item.trigger("mouseenter");
+        await item.trigger("mouseover");
+        await component.vm.$nextTick();
+        assert.strictEqual(component.emitted("update:highlighted").length, 1);
```
NFSS10, line 118: Move the cursor to other items, also check the `update:highlighted` event value

vue/test/components/dropdown/dropdown.test.js
```diff
@@ -0,0 +1,142 @@
...
+            {
+                value: "text_2"
+            }
+        ];
+        const component = base.getComponent("Dropdown", {
+            props: { items: items }
+        });
+
+        const numItems = items.length;
+        const itemIndex = Math.floor(Math.random() * numItems);
```
NFSS10, line 136: I think its better to use constant indexes to ensure the test will run the same way everytime

vue/test/components/dropdown/dropdown.test.js
```diff
@@ -0,0 +1,142 @@
...
+        const item = component.findAll(".dropdown-item").at(itemIndex);
+        await item.trigger("click");
+        await component.vm.$nextTick();
+        assert.strictEqual(component.emitted("item-clicked").length, 1);
```
NFSS10, line 140: Check the `item-clicked` event values

vue/test/components/dropdown/dropdown.test.js
```diff
@@ -0,0 +1,142 @@
...
+        const messageEmpty = "No items";
+        const component = base.getComponent("Dropdown", {
+            props: { items: items, messageEmpty: messageEmpty }
+        });
+
+        const dropdownItems = component.findAll(".dropdown-item").length;
+        const messageEmptyItem = component.findAll(".dropdown-item-empty").length;
+
+        assert.strictEqual(dropdownItems, 0);
+        assert.strictEqual(messageEmptyItem, 1);
```
NFSS10, line 16: Maybe change the names to represent quantity?

joamag:
Adding `fast-track` as we have a use case for this new component behavior here:

Capture

vue/test/components/dropdown/dropdown.test.js
```diff
@@ -0,0 +1,165 @@
...
+        assert.strictEqual(highlightedItem, highlightedItem);
+        await item1.trigger("mouseleave");
+
+        const itemIndex2 = 1;
+        const item2 = component.findAll(".dropdown-item").at(itemIndex2);
+        await item2.trigger("mouseenter");
+        await item2.trigger("mouseover");
+        highlightedItem = parseInt(Object.keys(component.emitted("update:highlighted")[2][0])[0]);
+        assert.strictEqual(component.emitted("update:highlighted").length, 3);
+        assert.strictEqual(highlightedItem, itemIndex2);
```
NFSS10, line 140: I think this test is too complex.
It would probably be better to make something like
```
component.vm.$on("update:highlighted", value => {
...
});
```
Where you listen for the event to be triggered, then do something like push the value to some array for example and then check if the values and order are what we are expecting

vue/test/components/dropdown/dropdown.test.js
```diff
@@ -0,0 +1,165 @@
...
+            props: { items: items }
+        });
+
+        const item = component.findAll(".dropdown-item").at(1);
+        await item.trigger("click");
+        const clickedItemValue = component.emitted("item-clicked")[0][0].value;
+        const clickedItemIndex = component.emitted("item-clicked")[0][1];
+        assert.strictEqual(component.emitted("item-clicked").length, 1);
+        assert.strictEqual(item.text(), clickedItemValue);
+        assert.strictEqual(1, clickedItemIndex);
```
NFSS10, line 163: Set a variable with the result of `component.emitted("item-clicked")` to reduce the `component.emitted` calls, like this:
`const clickedEvents = component.emitted("item-clicked");`

Also, in the asserts, do something like this:
`assert.deepEqual(clickedEvents[0], [{ value: 'text_2' }, 1]);`

vue/test/components/dropdown/dropdown.test.js
```diff
@@ -0,0 +1,165 @@
...
+        assert.strictEqual(component.emitted("update:highlighted").length, 1);
+        assert.strictEqual(highlightedItem, highlightedItem);
+        await item1.trigger("mouseleave");
+
+        const itemIndex2 = 1;
+        const item2 = component.findAll(".dropdown-item").at(itemIndex2);
+        await item2.trigger("mouseenter");
+        await item2.trigger("mouseover");
+        highlightedItem = parseInt(Object.keys(component.emitted("update:highlighted")[2][0])[0]);
+        assert.strictEqual(component.emitted("update:highlighted").length, 3);
```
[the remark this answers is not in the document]
Benedito0: @NFSS10 can you clarify ?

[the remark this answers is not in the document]
NFSS10: "Do you know why there are 3 events being emitted?"
Fixed, sry 😅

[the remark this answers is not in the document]
Benedito0: On mouseleave the highlighted gets updated too

vue/test/components/dropdown/dropdown.test.js
```diff
@@ -0,0 +1,164 @@
...
+            {
+                value: "text_2"
+            }
+        ];
+        const component = base.getComponent("Dropdown", {
+            props: { items: items }
+        });
+
+        const dropdownItems = component.findAll(".dropdown-item");
+        const itemsLength = dropdownItems.length;
```
NFSS10, line 33: Remove this line and use `dropdownItems.length` directly

vue/components/ui/atoms/dropdown/dropdown.vue
```diff
@@ -169,6 +178,10 @@ export const Dropdown = {
             type: String,
             default: "bottom"
         },
+        messageEmpty: {
+            type: String,
+            default: null
```
gcandal, line 183: Give it some kind of default.

vue/test/components/dropdown/dropdown.test.js
```diff
@@ -0,0 +1,163 @@
...
+            props: { items: items }
+        });
+
+        const dropdownItems = component.findAll(".dropdown-item");
+        const firstItem = dropdownItems.at(0);
+        const secondItem = dropdownItems.at(1);
+        assert.strictEqual(dropdownItems.length, 2);
+        assert.strictEqual(firstItem.text(), "Text 1");
+        assert.strictEqual(secondItem.text(), "text_2");
+        assert.strictEqual(component.vm.$props.items, items);
```
gcandal, line 38: No need for this line because it's not testing internal behavior. What's the use to know that a prop was passed? The only way for that to fail was that if `base.getComponent` was bugged.

vue/test/components/dropdown/dropdown.test.js
```diff
@@ -0,0 +1,163 @@
...
+        ];
+        const component = base.getComponent("Dropdown", {
+            props: { items: items }
+        });
+
+        const dropdownItems = component.findAll(".dropdown-item");
+        const firstItem = dropdownItems.at(0);
+        const secondItem = dropdownItems.at(1);
+        assert.strictEqual(dropdownItems.length, 2);
+        assert.strictEqual(firstItem.text(), "Text 1");
```
gcandal, line 36: Go for a one-liner: `assert.strictEqual(dropdownItems[0].text(), "Text 1");`

vue/test/components/dropdown/dropdown.test.js
```diff
@@ -0,0 +1,163 @@
...
+            }
+        ];
+        const component = base.getComponent("Dropdown", {
+            props: { items: items }
+        });
+
+        assert.strictEqual(component.vm.$data.visibleData, true);
+        const item = component.findAll(".dropdown-item").at(0);
+        await item.trigger("click");
+        assert.strictEqual(component.vm.$data.visibleData, false);
```
gcandal, line 58: Testing internal data has a place, but you also need to check external attributes (checking visibility, checking the text, etc).

gcandal: This applies to all tests: checking internal data is not enough.

gcandal:
@Benedito0 this also requires merging.

PR #343, opened 2020-07-01
Revert "Add semantic classes to dropdown items"

joamag:
Reverts ripe-tech/ripe-components-vue#342


joamag:
@gcandal reversed these changes as I believe we should use a custom `data-value` attribute instead of the class

PR #345, opened 2020-07-03
Add data-value to dropdown

gcandal:


joamag:
👍 let's back-port this solution...

PR #355, opened 2020-08-28
Support for breadcrumbs in container-ripe

NFSS10:
| - | - |
| --- | --- |
| Issue | -- |
| Dependencies | -- |
| Decisions | • Prop `title` in `container-ripe` and `details-ripe`component now accepts `array`'s so that we can choose between breadcrumbs and text <br>• Updated `container-ripe` and `details-ripe` storybook stories |
| Animated GIF | **`container-ripe`:**<br>container-use_breadcrumbs_when_title_is_array<br><br>**`details-ripe`:**<br>imagem |

### DIFF
--- a/vue/components/ui/atoms/container/container.vue
+++ b/vue/components/ui/atoms/container/container.vue
@@ -7,7 +7,7 @@
                     v-bind:font-size="breadcrumbsFontSize"
                     v-bind:separator="breadcrumbsSeparator"
                     v-bind:breadcrumbs="title"
-                    v-if="title && Array.isArray(title)"
+                    v-if="hasBreadcrumbs"
                 />
                 <title-ripe v-else-if="title">
                     {{ title }}
@@ -151,6 +151,9 @@ export const Container = {
         }
     },
     computed: {
+        hasBreadcrumbs() {
+            return this.title && Array.isArray(this.title);
+        },
         hasHeaderButtons() {
             return (
                 this.headerButtons.length > 0 ||
--- a/vue/components/ui/atoms/container/container.vue
+++ b/vue/components/ui/atoms/container/container.vue
@@ -3,9 +3,15 @@
         <div class="container-header" v-if="hasHeaderContent">
             <slot name="header-before" />
             <slot name="header">
-                <title-ripe v-if="title">
+                <title-ripe v-if="title && typeof title === 'string'">
                     {{ title }}
                 </title-ripe>
+                <breadcrumbs
+                    v-bind:font-size="breadcrumbsFontSize"
+                    v-bind:separator="breadcrumbsSeparator"
+                    v-bind:breadcrumbs="title"
+                    v-if="title && Array.isArray(title)"
+                />
                 <slot name="header-buttons-before" />
                 <slot name="header-buttons" v-if="hasHeaderButtons">
                     <div class="header-buttons">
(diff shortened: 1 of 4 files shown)


gcandal (changes requested):
No need to modify the Storybook. If we dump all the new enhancements there they'll become unmanageable. Their point is to broadly show what can a component do, not to showcase every kind of functionality.

Aren't there more components that use `container-ripe` that need their `title` changed to `type: String | Array`?

NFSS10:
> No need to modify the Storybook. If we dump all the new enhancements there they'll become unmanageable. Their point is to broadly show what can a component do, not to showcase every kind of functionality.
> 
> Aren't there more components that use `container-ripe` that need their `title` changed to `type: String | Array`?


Yes, `listing` and I kind hoped that you also asked about it. That one is a little weird as it doesn't use the prop `title`. It uses a prop `titleText`. It then builds the title text based on some rule that can or not use prefix. [See line 10](https://github.com/ripe-tech/ripe-components-vue/blob/5b4c85b594418d9f1c15be6b42e107380cf6d780/vue/components/ui/organisms/listing/listing.vue#L10
).

1- For this one we could have a prop breadcrums that if set would have priority over that rule
or
2- Probably the correct approach, change prop `titleText` -> `title`, accepts `String` and `Array` and then go to all other projects and update the component 😅

gcandal:
> > No need to modify the Storybook. If we dump all the new enhancements there they'll become unmanageable. Their point is to broadly show what can a component do, not to showcase every kind of functionality.
> > Aren't there more components that use `container-ripe` that need their `title` changed to `type: String | Array`?
> 
> Yes, `listing` and I kind hoped that you would asked about it. That one is a little weird as it doesn't use the prop `title`. It uses a prop `titleText`. It then builds the title text based on some rule that can or not use prefix. [See line 10](https://github.com/ripe-tech/ripe-components-vue/blob/5b4c85b594418d9f1c15be6b42e107380cf6d780/vue/components/ui/organisms/listing/listing.vue#L10).
> 
> 1- For this one we could have a prop `breadcrums` that if set would have priority over that rule
> or
> 2- Probably the correct approach, change prop `titleText` -> `title`, accepts `String` and `Array` and then go to all other projects and update the component 😅

Hm let's leave that one untouched, too much trouble for something we still don't have a use case for.

vue/components/ui/organisms/details/details.vue
```diff
@@ -9,7 +9,15 @@
...
-                    <title-ripe>{{ invalid ? invalidTitle : title }}</title-ripe>
+                    <breadcrumbs
+                        v-bind:font-size="breadcrumbsFontSize"
+                        v-bind:separator="breadcrumbsSeparator"
+                        v-bind:breadcrumbs="title"
+                        v-if="title && Array.isArray(title)"
+                    />
+                    <title-ripe v-else>
+                        {{ invalid ? invalidTitle : title }}
+                    </title-ripe>
```
gcandal, line 20: Apply this strategy "v-if=array + v-else" to `vue/components/ui/atoms/container/container.vue`

gcandal (approved):

vue/components/ui/atoms/container/container.vue
```diff
@@ -3,7 +3,13 @@
         <div class="container-header" v-if="hasHeaderContent">
             <slot name="header-before" />
             <slot name="header">
-                <title-ripe v-if="title">
+                <breadcrumbs
+                    v-bind:font-size="breadcrumbsFontSize"
+                    v-bind:separator="breadcrumbsSeparator"
+                    v-bind:breadcrumbs="title"
+                    v-if="title && Array.isArray(title)"
```
joamag, line 10: `hasBreadcrumbs` as computed

joamag (changes requested):

gcandal (approved):

joamag (approved):

PR #359, opened 2020-09-02
Fixed header-buttons aligment with search bar

NFSS10:
| - | - |
| --- | --- |
| Issue | -- |
| Dependencies | -- |
| Decisions | Search bar in `listing` wasn't aligned vertically with the header buttons, this PR tries to fix that |
| Animated GIF | **Before:**<br>imagem<br><br>**After fix:**<br>imagem |

### DIFF
--- a/vue/components/ui/organisms/listing/listing.vue
+++ b/vue/components/ui/organisms/listing/listing.vue
@@ -172,6 +172,10 @@ body.mobile .listing {
     min-height: 315px;
 }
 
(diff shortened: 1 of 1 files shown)


gcandal:
This alignment must be coherent with whatever alignment you use for the header buttons (when you add the capability to set them by props). Because of this, please do this when you add that new feature.

I understand why you tried doing so in a separate PR, but in this case this only makes sense in conjunction with the other change.

PR #361, opened 2020-09-02
Support for "button-color" in "container-ripe" header buttons

NFSS10:
| - | - |
| --- | --- |
| Issue | Request to add `button-color`s to listing (https://github.com/ripe-tech/ripe-registry-ui/pull/3) |
| Dependencies | https://github.com/ripe-tech/ripe-components-vue/pull/359 |
| Decisions | Support for `button-color`'s in header buttons. To use `button-color` instead of the default `button-icon`, set the property `isButtonColor` to `true`. The `button-color` props are supported via the `buttonColorProps`  |
| Animated GIF | container-support_for_button-color_in_header_buttons |

### DIFF
--- a/vue/components/ui/atoms/container/container.vue
+++ b/vue/components/ui/atoms/container/container.vue
@@ -21,7 +21,6 @@
                                 class="header-button"
                                 v-bind="{
                                     text: button.text,
-                                    icon: button.icon,
                                     size: 'small',
                                     minWidth: 0,
                                     ...button.buttonColorProps
--- a/vue/components/ui/atoms/container/container.vue
+++ b/vue/components/ui/atoms/container/container.vue
@@ -120,7 +120,8 @@ body.mobile .container-ripe > .container-header {
 
 .container-ripe > .container-header > .header-buttons > .header-button {
     margin: 0px 4px 0px 0px;
-}
+    vertical-align: middle;
+ }
 
 body.tablet .container-ripe > .container-header > .header-buttons,
 body.mobile .container-ripe > .container-header > .header-buttons {
--- a/vue/components/ui/atoms/container/container.vue
+++ b/vue/components/ui/atoms/container/container.vue
@@ -200,6 +200,7 @@ export const Container = {
     },
     methods: {
         onHeaderButtonClick(event, buttonId) {
+            this.$emit(`header-button:click:${buttonId}`, event);
             this.$emit("header-button:click", event, buttonId);
         }
     }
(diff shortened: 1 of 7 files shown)


vue/components/ui/atoms/container/container.vue
```diff
@@ -16,30 +16,44 @@
...
+                            <button-color
+                                class="header-button"
+                                v-bind="{
+                                    text: button.text,
+                                    size: 'small',
+                                    minWidth: 0,
+                                    ...button.buttonColorProps
+                                }"
+                                v-show="!button.hide"
+                                v-if="button.isButtonColor"
```
gcandal, line 29: `button.type === 'color'` instead

gcandal (changes requested):

vue/components/ui/atoms/container/container.vue
NFSS10: Ups, forgot that :(

vue/components/ui/atoms/container/container.vue
```diff
@@ -104,6 +118,11 @@ body.mobile .container-ripe > .container-header {
     user-select: none;
 }
 
+.container-ripe > .container-header > .header-buttons > .header-button {
+    margin: 0px 4px 0px 0px;
```
gcandal, line 122: Shouldn't this be applied to button-colors only?

vue/components/ui/atoms/container/container.vue
```diff
@@ -180,7 +199,7 @@ export const Container = {
         }
     },
     methods: {
-        onButtonIconClick(event, buttonId) {
+        onHeaderButtonClick(event, buttonId) {
```
gcandal, line 202: Maybe emit `header-button:click:buttonId` too, so then you can consume it with `v-on:header-button:click:add`.

vue/components/ui/atoms/container/container.vue
NFSS10: Both `button-icon`s and `button-color`s don't have any margin between them. I guess if we wanted to add margin only to `button-color`s then yes, I need to change
`.container-ripe > .container-header > .header-buttons > .header-button {` 
to
`.container-ripe > .container-header > .header-buttons > .header-button.button-color {`

gcandal: I know they don't, but if button-icons didn't have a margin before this change, they shouldn't gain a margin now without very good reason. They don't have a margin because they only occupy the full space when they are `:hover`, so they don't need it.

Long story short: please make the change you mentioned.

gcandal (changes requested):

gcandal (approved):

joamag:
@gcandal we need a better use case if we want to add this feature set

PR #370, opened 2020-09-15
Remove padding when text is null

Benedito0:
| - | - |
| --- | --- |
| Decisions |The need for this came from working on this [issue](https://github.com/ripe-tech/ripe-pulse/issues/146) on Pulse, noticing the disalignment of icon in the button. Keeping the padding when there is a slot because most usages of this component use the slot with just text. |

### DIFF
--- a/vue/components/ui/atoms/button-color/button-color.vue
+++ b/vue/components/ui/atoms/button-color/button-color.vue
@@ -18,7 +18,6 @@
             v-bind:color="iconColor"
             v-bind:width="iconSize"
             v-bind:height="iconSize"
-            v-bind:style="iconStyle"
             v-if="icon && !loading"
         />
         <icon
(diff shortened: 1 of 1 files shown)


gcandal (approved):

vue/components/ui/atoms/button-color/button-color.vue
```diff
@@ -406,6 +407,11 @@ export const ButtonColor = {
             }
```
joamag, line 408: add a `no-text` class for when there's no text and use it to invalidate the padding

joamag (changes requested):

joamag (approved):

joamag (approved):

joamag:
@gcandal I believe this requires back-port

gcandal:
> @gcandal I believe this requires back-port

@Benedito0 please take care of this.

joamag:
@gcandal it seems that there's no need for back-porting

PR #371, opened 2020-09-15
Remove babel polyfill

gcandal:
| - | - |
| --- | --- |
| Dependencies | https://github.com/hivesolutions/uxf_webpack/pull/1 |


joamag:
@gcandal First we need to prove that this works before propagation

PR #374, opened 2020-09-18
Generic form component

gcandal:
| - | - |
| --- | --- |
| Animated GIF |  |

### DIFF
--- a/vue/components/ui/atoms/container/container.vue
+++ b/vue/components/ui/atoms/container/container.vue
@@ -162,6 +162,7 @@ export const Container = {
     methods: {
         onButtonIconClick(event, buttonId) {
             this.$emit("header-button:click", event, buttonId);
+            this.$emit(`header-button:click:${buttonId}`, event);
         }
     }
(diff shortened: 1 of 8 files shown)


joamag (approved):

joamag:
@gcandal it's critical that we make this form way more flexible allowing the options to add more buttons at the buttons groups, etc. 
The law is the same... easy usage by default but make it powerful enough by adding extensibility support

joamag:
@gcandal by the way that loading indicator inside the save button looks a bit crappy the min-width should safe-guard the button

joamag:
I've reopened the branch by reversing the commit as there are many things that need to be fixed:
Some examples by comparing RIPE White admin (correct) with the current implementation:

Notice:
1. Size of the buttons
2. Alignment of the columns
3. ... and other small details

Capture3
Capture2

joamag:
The new PR is here https://github.com/ripe-tech/ripe-components-vue/pull/378

PR #379, opened 2020-09-22
Support select slots in form-ripe

NFSS10:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-registry-ui/pull/16 |
| Dependencies | -- |
| Decisions | Need to use the select slots to be able to customize the select values UI<br>Example of slots in use below, take notice in the two selects that have their entries with an avatar, name and email|
| Animated GIF | <br>client-create |

### DIFF
--- a/vue/components/ui/organisms/form/form.vue
+++ b/vue/components/ui/organisms/form/form.vue
@@ -46,7 +46,10 @@
                                             <slot v-bind:name="`select-${field.value}-selected`" />
                                         </template>
                                         <template v-slot="{ item }">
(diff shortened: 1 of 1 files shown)


NFSS10:
Need to check things

joamag:
@NFSS10 Please elaborate on the use-case of this change... use a screenshot if necessary

NFSS10:
> @NFSS10 Please elaborate on the use-case of this change... use a screenshot if necessary

Updated decisions field and added an example of client create, where we have selects with custom entries (image, name and email)

joamag (approved):

PR #378, opened 2020-09-21
Generic Form component - second try

joamag:
| - | - |
| --- | --- |
| Issue |The original PR (conversation) here https://github.com/ripe-tech/ripe-components-vue/pull/374 |
| Dependencies | -- |
| Decisions | `form-ripe` component |
| Animated GIF | generic-form_fixes |

PR #384, opened 2020-10-01
Field types improvements

NFSS10:
| - | - |
| --- | --- |
| Issue |https://github.com/ripe-tech/ripe-components-vue/pull/378#issuecomment-701349578 |
| Dependencies | -- |
| Decisions | • Can use `meta` to specify the input type <br>• To use textarea, instead of just specifying type "textarea" we now do type "text" and  meta "long" <br>• Added new image url input type |
| Screenshot | imagem |

### DIFF
--- a/vue/components/ui/organisms/form/form.stories.js
+++ b/vue/components/ui/organisms/form/form.stories.js
@@ -19,11 +19,12 @@ storiesOf("Organisms", module)
                                 fields: [
                                     {
                                         value: "username",
-                                        type: "string"
+                                        type: "text"
                                     },
                                     {
                                         value: "email",
-                                        type: "email"
+                                        type: "text",
+                                        meta: "email"
                                     }
                                 ]
                             },
@@ -32,11 +33,12 @@ storiesOf("Organisms", module)
                                 fields: [
                                     {
                                         value: "username2",
-                                        type: "string"
+                                        type: "text"
                                     },
                                     {
                                         value: "email2",
-                                        type: "email"
+                                        type: "text",
+                                        meta: "email"
                                     }
                                 ]
                             }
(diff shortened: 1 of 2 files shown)


vue/components/ui/organisms/form/form.vue
```diff
@@ -27,16 +27,34 @@
                                     v-bind:key="field.value"
                                 >
                                     <input-ripe
-                                        v-bind:type="field.type"
+                                        v-bind:type="field.meta"
```
gcandal, line 30: Make the mapping in a separate function, even if for now they're the same.

vue/components/ui/organisms/form/form.vue
```diff
@@ -27,16 +27,34 @@
...
                                     <input-ripe
-                                        v-bind:type="field.type"
+                                        v-bind:type="field.meta"
                                         v-bind="field.props"
                                         v-bind:value="values[field.value]"
                                         v-if="
-                                            ['string', 'number', 'email', 'input'].includes(
-                                                field.type
-                                            )
+                                            field.type === 'text' &&
```
gcandal, line 34: Let's simplify this by using v-else-if and placing the more specific inputs first.

vue/components/ui/organisms/form/form.stories.js
```diff
@@ -47,11 +49,17 @@ storiesOf("Organisms", module)
...
                                         value: "username3",
-                                        type: "string"
+                                        type: "text"
                                     },
                                     {
-                                        value: "email3",
-                                        type: "email"
+                                        value: "description",
+                                        type: "text",
+                                        meta: "long"
```
gcandal, line 57: Let's use `longtext` in order to be aligned with appier: https://github.com/hivesolutions/appier/blob/0d1b84a90fccd6759b38fc643012e85281ed7d29/src/appier/model.py#L81

vue/components/ui/organisms/form/form.vue
```diff
@@ -147,6 +159,18 @@ body.mobile .form > .form-form > .tabs .column {
 .form > .form-form > .tabs .column > .section > .section-field:last-child {
     margin-bottom: 0px;
 }
+
+.form > .form-form > .tabs .column > .section > .section-field ::v-deep .content {
```
gcandal, line 163: Which is the component whose content alignment you're trying to override? Shouldn't it be a prop of that component and not a v-deep?

gcandal (changes requested):

PR #385, opened 2020-10-01
Added semantic classes

NFSS10:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-components-vue/pull/378#issuecomment-702214987 |
| Dependencies | -- |
| Decisions | • Added `buildSlug` method <br>• Added semantic classes to column, section and form input <br>• Improved tabs classes |
| Animated GIF | -- |

### DIFF
--- a/js/util.js
+++ b/js/util.js
@@ -2,4 +2,12 @@ export const normalize = value => {
     return value.split("_").join(" ");
 };
 
-export default normalize;
+export const buildslug = value => {
+    return value
+        .toLowerCase()
+        .normalize("NFD")
+        .trim()
+        .replace(/\s+/g, "-")
+        .replace(/[^\w\-]+/g, "")
+        .replace(/\-\-+/g, "-");
+};
--- a/js/util.js
+++ b/js/util.js
@@ -5,9 +5,8 @@ export const normalize = value => {
 export const buildslug = value => {
     return value
         .toLowerCase()
-        .normalize("NFD")
         .trim()
-        .replace(/\s+/g, "-")
-        .replace(/[^\w\-]+/g, "")
-        .replace(/\-\-+/g, "-");
+        .replace(/\s+/g, "-") // replace spaces with -
+        .replace(/[^\w\-]+/g, "") // remove all non-word chars
+        .replace(/\-\-+/g, "-"); // replace multiple - with a single -
 };
--- a/vue/components/ui/molecules/tabs/tabs.vue
+++ b/vue/components/ui/molecules/tabs/tabs.vue
@@ -152,6 +152,7 @@ export const Tabs = {
         tabClasses(tab) {
             const base = { tab: true };
             base[`tab-${tab.value}`] = true;
+            base[`tab-${tab.value.replace(" ", "-").toLowerCase()}`] = true;
             return base;
         },
         onEnter(index) {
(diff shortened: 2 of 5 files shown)


vue/components/ui/molecules/tabs/tabs.vue
```diff
@@ -152,6 +152,7 @@ export const Tabs = {
         tabClasses(tab) {
             const base = { tab: true };
             base[`tab-${tab.value}`] = true;
+            base[`tab-${tab.value.replace(" ", "-").toLowerCase()}`] = true;
```
gcandal, line 155: Use https://github.com/ripe-tech/ripe-components-vue/blob/master/js/util.js#L1

vue/components/ui/organisms/form/form.vue
```diff
@@ -299,6 +309,16 @@ export const Form = {
             };
             return base;
         },
+        sectionClasses(index) {
+            return `section-${index}`;
```
gcandal, line 313: Use `section.title` (same as v-bind:key)

vue/components/ui/organisms/form/form.vue
```diff
@@ -299,6 +309,16 @@ export const Form = {
...
         },
+        sectionClasses(index) {
+            return `section-${index}`;
+        },
+        formInputClasses(field) {
+            const base = {};
+            base[`field-${field.value}`] = Boolean(field.value);
+            base[`field-${field.type}`] = Boolean(field.type);
+            base[`field-${field.meta}`] = Boolean(field.meta);
+            return base;
```
gcandal, line 320: Let's go with the same old approach where possible

```suggestion
            const base = {};
            base[`field-${field.value}`] = true;
            base[`field-${field.type}`] = true;
            base[`field-${field.meta}`] = Boolean(field.meta);
            return base;

```

gcandal (changes requested):

vue/components/ui/molecules/tabs/tabs.vue
NFSS10: I made this change so that we have improved classes, normalize wouldn't solve this example:
imagem
would become:
imagem

Taking a better look, we should probably do something like this in for section classes too and use something that handles invalid characters too

vue/components/ui/organisms/form/form.vue
NFSS10: https://github.com/ripe-tech/ripe-components-vue/pull/385#discussion_r498660431

vue/components/ui/molecules/tabs/tabs.vue
gcandal: Move this to a `buildSlug` function inside `util`

NFSS10: Done, for example, "Column #1 Section #1" now parses to "column-1-section-1"

js/util.js
```diff
@@ -2,4 +2,11 @@ export const normalize = value => {
...
 };
 
-export default normalize;
+export const buildslug = value => {
+    return value
+        .toLowerCase()
+        .trim()
+        .replace(/\s+/g, "-") // replace spaces with -
+        .replace(/[^\w\-]+/g, "") // remove all non-word chars
+        .replace(/\-\-+/g, "-"); // replace multiple - with a single -
```
NFSS10, line 11: Comments upper line

Issue #408, opened 2020-12-09
Steps


gcandal:
## Description

**Base issue**: https://github.com/ripe-tech/ripe-core/issues/4425

Create a `steps` component that represents where a user is in a given context. Props:
- steps (list of strings)
- step (number representing the index of the current step)

Visual:

![image](https://user-images.githubusercontent.com/3048875/101664693-0e7a1100-3a44-11eb-8671-44432920c763.png)

PR #414, opened 2020-12-17
#408 Steps

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-components-vue/issues/408 |
| Decisions | Add simple steps component with tests and docs. Define a structure `step > step-number` + `step > step-text` with flex to center the labels on top of the progress-bar, this was the cleanest approach. Add clickable steps with prop default to false. |
| Animated GIF | Peek 2020-12-17 14-57 |


vue/components/ui/molecules/steps/steps.stories.js
```diff
@@ -0,0 +1,64 @@
...
+            onProgressBarSimulationButtonClick() {
+                if (this.progressTimer) clearTimeout(this.progressTimer);
+                this.simulatedCurrentStep = 0;
+                this.updateProgress();
+            },
+            updateProgress() {
+                if (this.simulatedCurrentStep >= 5) return;
+                this.simulatedCurrentStep += this.simulationAdvanceStep;
+                this.simulatedCurrentStep = Math.min(this.simulatedCurrentStep, 5);
+                this.progressTimer = setTimeout(this.updateProgress, this.simulationStepTimeMs);
```
gcandal, line 43: Don't set the timeout when you're already at max.

vue/test/components/steps/steps.test.js
```diff
@@ -0,0 +1,75 @@
...
+const base = require("../../base");
+
+describe("Steps", () => {
+    it("should initialize the steps", () => {
+        const component = base.getComponent("Steps", {
+            props: { steps: ["A", "B", "C"], step: 1 }
+        });
+        assert.strictEqual(component.exists(), true);
+    });
+    it("should only highlight the current step", () => {
```
gcandal, line 11: Newline between `it`s I think. Did you get this (they being together) from any place?

vue/test/components/steps/steps.test.js
```diff
@@ -0,0 +1,75 @@
...
+            props: { steps: ["A", "B", "C", "D", "E"], step: 2 }
+        });
+        const steps = component.findAll(".step .highlight");
+        assert.strictEqual(steps.at(0).isVisible(), false);
+        assert.strictEqual(steps.at(1).isVisible(), true);
+        assert.strictEqual(steps.at(2).isVisible(), false);
+        assert.strictEqual(steps.at(2).isVisible(), false);
+        assert.strictEqual(steps.at(2).isVisible(), false);
+    });
+    it("should change to clicked step and emit change", async () => {
```
gcandal, line 22: Check that the highlight changed.

vue/test/components/steps/steps.test.js
```diff
@@ -0,0 +1,75 @@
...
+        const component = base.getComponent("Steps", {
+            props: { steps: ["A", "B", "C"], step: 1 }
+        });
+        assert.strictEqual(component.exists(), true);
+    });
+    it("should only highlight the current step", () => {
+        const component = base.getComponent("Steps", {
+            props: { steps: ["A", "B", "C", "D", "E"], step: 2 }
+        });
+        const steps = component.findAll(".step .highlight");
```
gcandal, line 15: Check that all are visible, even if not highlighted.

vue/test/components/steps/steps.test.js
```diff
@@ -0,0 +1,75 @@
...
+    it("should have white background if step is not completed and gray if otherwise", () => {
+        const component = base.getComponent("Steps", {
+            props: { steps: ["A", "B", "C", "D", "E"], step: 3 }
+        });
+        const steps = component.findAll(".step .step-number");
+        assert.strictEqual(
+            steps
+                .at(0)
+                .attributes("style")
+                .includes("rgb(89, 98, 109)"),
```
gcandal, line 43: I understand the temptation to use this, but it seems fragile. Use a class to set the color and make the test use that class. That way you got an abstract definition that is not prone to break when you make UI changes.

vue/test/components/steps/steps.test.js
gcandal, line 22: Check you can't do this when not clickable

gcandal: Asserting internal data is like documenting internal methods: even if it's great when you do it and it helps a lot on future developments, it is neither absolutely necessary nor enough.

What I mean is that unit tests usually treat components under test as black boxes: you don't care how they're implemented, you just care of what you can see outside their interface (UI, method calls, events, side-effects like writing to file, etc).

gcandal (changes requested):
This looks cool.

Add `cursor: pointer` when `clickable`.

vue/components/ui/molecules/steps/steps.vue
```diff
@@ -0,0 +1,163 @@
...
+            default: false
+        }
+    },
+    data: function() {
+        return {
+            stepData: this.step
+        };
+    },
+    computed: {
+        stepClasses(index) {
```
gcandal, line 134: Move this to `methods` and get rid of the lambda, I think it will still be reactive. At least, we're doing it in a ton of places.

In any case, if you wanted this as `computed` you couldn't pass arguments to it. In your case, the argument is being ignored, it's the lambda that is capturing the index.

gcandal (changes requested):

vue/components/ui/molecules/steps/steps.vue
3rdvision: If I move it to methods, it is not reactive and doesn't seem to work.

gcandal: Just change stepClasses(index) to stepClasses then, that should work.

gcandal (approved):

joamag (approved):

PR #424, opened 2021-01-12
Blob mixin

JoaoLeal12:
|-|-|
|---|---|
|Issue| https://github.com/ripe-tech/ripe-util-vue/pull/9#discussion_r555029391|
|Description|Copied code from https://github.com/ripe-tech/ripe-white/blob/b6a0ca6d79788a7af8968e601f4270012b331b65/js/plugins/technical/download/download.js#L112|


vue/mixins/blob.js
```diff
@@ -0,0 +1,22 @@
+export const blobMixin = {
+    methods: {
+        /**
+         * Downloads a blob file with a given filename
+         *
+         * @param {Object} blob File to be downloaded
+         * @param {String} filename Name to give the file
+         */
+        downloadBlob(blob, filename) {
```
gcandal, line 9: This can actually be used with `File`, `Blob` or `MediaSource` so just name it `download`

gcandal: Because of this, rename to `downloadMixin` as well.

See: https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL

vue/mixins/blob.js
```diff
@@ -0,0 +1,22 @@
+export const blobMixin = {
+    methods: {
+        /**
+         * Downloads a blob file with a given filename
+         *
+         * @param {Object} blob File to be downloaded
+         * @param {String} filename Name to give the file
```
gcandal, line 7: Period at the end of sentences

gcandal (changes requested):

gcandal:
@JoaoLeal12 please include in the description where was this copied from (ripe-white something)

gcandal (approved):
@JoaoLeal12 I'm accepting but next time please use the PR template.

joamag (approved):

PR #426, opened 2021-01-13
feat: extended icons via $iconSources

gcandal:
| - | - |
| --- | --- |
| Decisions | Useful to allow extended item-set for client apps. |


joamag:
@gcandal I believe we need to back-port this (white?)

joamag (approved):

PR #428, opened 2021-01-18
fix: chat top margin

gcandal:
| - | - |
| --- | --- |
| Now | image |
| Before | image |


joamag:
I'm not sure this is really a bug...

gcandal:
Not sure either but it seems to look better, the previous UI felt weird. It's also aligned with [zeplin](https://zpl.io/2vo0rxe):

image

gcandal:
Delaying this

PR #438, opened 2021-02-04
Fix select-ripe value selection error in a mobile device

NFSS10:
| - | - |
| --- | --- |
| Issue | Error when selecting a value in a mobile device |
| Dependencies | -- |
| Decisions | Fixed error when trying to select a value in a mobile device |
| Screenshot | **Error screenshot:**<br>imagem |

### DIFF
--- a/vue/components/ui/molecules/select/select.vue
+++ b/vue/components/ui/molecules/select/select.vue
@@ -5,7 +5,7 @@
             class="dropdown-select"
             v-bind:value="value"
             v-if="isDevice()"
-            v-on:change="onSelectChange($event.target.value)"
+            v-on:change="event => onSelectChange(event.target.value)"
         >
             <option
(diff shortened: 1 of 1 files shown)


vue/components/ui/molecules/select/select.vue
```diff
@@ -5,7 +5,7 @@
             class="dropdown-select"
             v-bind:value="value"
             v-if="isDevice()"
-            v-on:change="() => onSelectChange($event.target.value)"
+            v-on:change="onSelectChange($event.target.value)"
```
gcandal, line 8: Standard: don't rely on Vue callback transformation. When binding to callbacks pass a function.

```suggestion
            v-on:change="event => onSelectChange(event.target.value)"
```

joamag (approved):

PR #442, opened 2021-02-19
feat: set of improvements

gcandal:
| - | - |
| --- | --- |
| Issue | Order import form. |
| Decisions | Includes: attachments click handler and optional hover, pretty, and file conversion and tab opening utils. |


vue/mixins/url.js
```diff
@@ -0,0 +1,11 @@
+export const urlMixin = {
+    methods: {
+        openNewTab(url) {
+            const anchor = Object.assign(document.createElement("a"), {
```
joamag, line 4: where's this snippet coming from? Why not window.location or window.open?

joamag (changes requested):

joamag (approved):

PR #452, opened 2021-03-04
Select: Mobile native toggle

3rdvision:
| - | - |
| --- | --- |
| Issue | There is a double arrow problem in `<select-twitch>` which happens when we have a mobile user agent. |
| Decisions | Hide `background-image` when using native select . |
| Animated GIF | The double arrow bug. <br> image |

### DIFF
--- a/vue/components/ui/molecules/select/select.vue
+++ b/vue/components/ui/molecules/select/select.vue
@@ -213,7 +213,7 @@ export const Select = {
         },
         mobileNative: {
             type: Boolean,
-            default: true
+            default: false
         }
     },
     data: function() {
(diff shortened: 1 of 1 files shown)


gcandal:
@3rdvision this actually should use the same solution as the ripe-white select: https://github.com/ripe-tech/ripe-white/blob/56526081ba1e9ce8e8aff6f2d5b82bcd55b81f4f/vue/components/molecules/select/select.vue#L125

I just realized, after seeing this PR, that we had already fixed it in the past. Guess we forgot to backport it here.

3rdvision:
> ripe-white sel

I think the CSS change could still be desired as I'm not sure we want this:

image
(double arrow when mobile agent)

gcandal (approved):

joamag (approved):

PR #453, opened 2021-03-05
fix: tooltip appears even with empty text

gcandal:


joamag (approved):

joamag:
@gcandal we need to back-port this one

PR #454, opened 2021-03-05
Select: fix text vertical alignment

3rdvision:
| - | - |
| --- | --- |
| Issue | Select has slight vertical misalignment <br> image |
| Decisions | In order to vertical align using `line-height`, it should have the same amount of height as the container it is in. The container has 34px but 4px of those are for the border so in reality it has 30px. |
| Animated GIF | image |

### DIFF
--- a/vue/components/ui/molecules/select/select.vue
+++ b/vue/components/ui/molecules/select/select.vue
@@ -91,7 +91,7 @@
     font-size: 13px;
     height: 34px;
     letter-spacing: 0.25px;
-    line-height: 34px;
+    line-height: 30px;
     outline: none;
     overflow: hidden;
     padding-left: 12px;


gcandal (approved):

joamag:
This is not correct, please note that this breaks the UI of the default select in ripe components. Probably a box-sizing-related issue.

image

@gcandal @3rdvision let's be careful about these changes and post screens of the base ripe components Vue component instead of their app-specific applications (which may override some settings)

3rdvision:
> This is not correct, please note that this breaks the UI of the default select in ripe components. Probably a box-sizing-related issue.
> 
> image
> 
> @gcandal @3rdvision let's be careful about these changes and post screens of the base ripe components Vue component instead of their app-specific applications (which may override some settings)

@joamag You're right as in this case `select-ripe` is only 32px of height. It's `select-twitch` that is 34px because it has a 2px border.

I still think `select-ripe` of components-vue has a slight 2px vertical misalignment.

I confirmed this with flex's vertical alignment and a virtual ruler:

Peek 2021-03-08 09-37


We could remove `box-sizing: border-box` which will increase `select-ripe` height to 36px:

Peek 2021-03-08 09-39

This may not be the best solution as other inputs all seem to have a `height: 34px` and a  `box-sizing: border-box`.
I think the optimal minimal solution would be to adjust the line-height to match the select's content-box which is 32 pixels and not 34 as it is now.

PR #457, opened 2021-03-10
feat: switcher text and color props

BeeMargarida:
| - | - |
| --- | --- |
| Issue | Related to https://github.com/ripe-tech/ripe-twitch/issues/42 |
| Dependencies | -- |
| Decisions | Switcher support for text on checked and unchecked status, and checked color. <br><br> Necessary due to the design in `twitch-admin-ui`, such as the `edit` and `creation`pages: <br>image <br> image|
| Animated GIF |switcher-textswitcher-use-case|

### DIFF
--- a/vue/components/ui/atoms/switcher/switcher.vue
+++ b/vue/components/ui/atoms/switcher/switcher.vue
@@ -1,6 +1,6 @@
 <template>
     <div class="switcher" v-bind:class="classes">
-        <div class="switcher-button-container" v-bind:style="styleContainer" v-on:click="onClick">
+        <div class="switcher-button-container" v-on:click="onClick">
             <div class="switcher-button" v-bind:style="toggleAnimation" />
         </div>
         <div class="switcher-text" v-if="text">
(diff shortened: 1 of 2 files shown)


gcandal:
Can't you use an approach similar to https://github.com/ripe-tech/ripe-twitch-ui/blob/master/components/atoms/button-twitch/button-twitch.vue ? Namely: using this component in a switcher-twitch and then v-deep everything?

gcandal:
The issue with using props to set the style is that will always be limited. If it was only one or two colors it would be ok but this commit includes CSS too, which is always risky.

However, with "v-deep inheritance", the sky is the limit, so in this particular instance let's go with it. In the future, it might make sense to add more props

gcandal (approved):

joamag:
Need two things:
* Where is this going to be used (different colors)
* Good rationale for the text as the switch component does not include text as of now

BeeMargarida:
> Need two things:
> 
>     * Where is this going to be used (different colors)
> 
>     * Good rationale for the text as the switch component does not include text as of now

Added to the PR description. Its necessary for the new designs of twitch-admin-ui where its necessary not only to switch the color, but also to show the meaning of both states of the switcher with the text on the right.

joamag:
> > Need two things:
> > ```
> > * Where is this going to be used (different colors)
> > 
> > * Good rationale for the text as the switch component does not include text as of now
> > ```
> 
> Added to the PR description. Its necessary for the new designs of twitch-admin-ui where its necessary not only to switch the color, but also to show the meaning of both states of the switcher with the text on the right.

We're not moving forward with those designs

vue/components/ui/atoms/switcher/switcher.vue
```diff
@@ -1,36 +1,51 @@
...
+    cursor: default;
+    opacity: 0.3;
+}
+
+.switcher > .switcher-button-container {
+    border: 2px solid;
     border-radius: 500px 500px 500px 500px;
     cursor: pointer;
     height: 20px;
+    position: relative;
```
joamag, line 30: remove the position relative

joamag (changes requested):

joamag:
Just talked with @gcandal we'll move with part of the changes defined extended the component scope

gcandal:
@joamag done.

joamag (approved):

joamag:
@gcandal we need to back-port this. Note that the grey version should be the default to one to ripe white

PR #458, opened 2021-03-11
feat: allow form components override

BeeMargarida:
| - | - |
| --- | --- |
| Issue | Related to https://github.com/ripe-tech/ripe-twitch/issues/84 |
| Dependencies | -- |
| Decisions | Code necessary to allow override form components with custom ones. <br><br> Its necessary to override the components of a form if a new more specific is needed (for example in the twitch-admin, instead of a switcher, its necessary to use a switcher with a specific behaviour). For that to be possible, a `slot` was added to the `form-input` value and the code necessary to override that slot was added to the container-form, so that its possible to use that slot when using the `entity-create` component. |
| Animated GIF | To allow something like <br> image|

### DIFF
--- a/vue/components/ui/organisms/container-form/container-form.vue
+++ b/vue/components/ui/organisms/container-form/container-form.vue
@@ -6,7 +6,16 @@
         v-on:header-button:click="onHeaderButtonClick"
         v-on:header-button:click:delete="onDeleteClick"
     >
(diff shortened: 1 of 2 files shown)


gcandal (approved):

joamag:
@BeeMargarida Please elaborate on the changes made and the purpose

BeeMargarida:
> @BeeMargarida Please elaborate on the changes made and the purpose

Updated the PR description with the reasoning

joamag (approved):

PR #462, opened 2021-03-12
feat: entity-create discard callback

BeeMargarida:
| - | - |
| --- | --- |
| Issue | Originated from https://github.com/ripe-tech/ripe-twitch/issues/84 |
| Dependencies | -- |
| Decisions | In some cases (twitch-admin for example), the discard button can have other functionality (and be called `Cancel` or other things instead of `Discard`). In twitch-admin, the button is called `Cancel` and when clicked, it should redirect to the listing page. Currently, the button only cleans the values of the form and there is no way of doing additional logic.<br> A prop was added called `onDiscardCallback` that when defined is called when the discard button is clicked. |
| Animated GIF | -- |

### DIFF
--- a/vue/components/ui/organisms/entity-create/entity-create.vue
+++ b/vue/components/ui/organisms/entity-create/entity-create.vue
@@ -83,6 +83,13 @@ export const EntityCreate = {
             default: null
         },
         /**
+         * Method to be called on discard.
+         */
+        onDiscardCallback: {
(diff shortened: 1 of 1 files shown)


gcandal:
Can't you just override `onDiscard` via `formProps`? Example:

https://github.com/ripe-tech/ripe-white-admin-ui/pull/36/files

PR #464, opened 2021-03-18
feat: added slot to listing's header button

BeeMargarida:
| - | - |
| --- | --- |
| Issue | Related to https://github.com/ripe-tech/ripe-twitch/issues/83|
| Dependencies | -- |
| Decisions | This slot is necessary for overriding the create entity button in the listing's header. For example, in this screen below, the button needs to have a different text and icon. In addition to this, the button opens a modal instead of redirecting to another page. image <br> To allow that kind of behaviour, the easier solution was to create a slot, which changes nothing in the overall behaviour of the component. |
| Animated GIF | -- |

### DIFF
--- a/vue/components/ui/organisms/listing/listing.vue
+++ b/vue/components/ui/organisms/listing/listing.vue
@@ -51,22 +51,24 @@
             </template>
             <template v-slot:header-buttons-inside-after>
                 <slot v-bind:name="'header-buttons-inside-after'" />
-                <router-link
-                    class="button-create"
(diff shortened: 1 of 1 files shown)


gcandal (approved):

joamag:
Why not use `headers-button-inside-after` and not set create URL?

Screenshot 2021-03-18 at 20 05 11

joamag:
but still I'm merging this one as I think it may make sense to have that surrounded by a slot

PR #469, opened 2021-03-25
fix: override icon size and margin-top to fix blurry icon

BeeMargarida:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-twitch/issues/84#issuecomment-806551523 |
| Dependencies | |
| Decisions | Although the fix talked about was changing the width/height of the icon by prop, it would not change, since `button-color` is setting the size of the icon using css (18px). Using a prop in `button-color` to control the width and height of the icon and then using a computed style, it would still be necessary to also add another prop to control the icon margin top (its dependent on the size). <br> This would be a big change only for this specific fix. The solution here presented simply overrides the style in listing, where the icon is already pre-defined. |
| Animated GIF | **Before**:<br>image <br>**Now**: image |

### DIFF
--- a/vue/components/ui/organisms/listing/listing.vue
+++ b/vue/components/ui/organisms/listing/listing.vue
@@ -63,7 +63,6 @@
                             v-bind:size="'small'"
                             v-bind:alignment="'left'"
                             v-bind:icon="'add'"
-                            v-bind:icon-size="17"
                             v-bind:min-width="0"
                             v-bind:href="href"
                             v-on:click="navigate"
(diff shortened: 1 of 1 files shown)


gcandal:
Shouldn't this be on entity-list?

BeeMargarida:
> Shouldn't this be on entity-list?

The listing is the component that "sets" the button-create already, the entity-list uses the listing and adds the action buttons to each table row.

gcandal:
Closing as we don't have a good grasp on what's going on:
- Only some browsers / devices
- Not all icons have the same fix
- Couldn't reproduce the problem by using size 17 direcly in ikonate
- Ikonate can't export at 17

PR #470, opened 2021-03-26
fix: highlight color and highlight timing

BeeMargarida:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-twitch/issues/42#issuecomment-806548905 |
| Dependencies | -- |
| Decisions | Fix highlight color, to make it match the design (the previous color was the color if the hover effect is active, giving it less opacity). Also fixed the default timings, making them match the timings given to the, for example, the notification timeout in entity-show. The original timing was a bit too long. |
| Testing | https://ripe-components-vue-git-ms-fix-image-item-highlight-platforme.vercel.app/?path=/story/components-atoms-image-item--image-item |
| Animated GIF |fix-highlight-story<br>fix-highlight|

### DIFF
--- a/vue/components/ui/atoms/image-item/image-item.vue
+++ b/vue/components/ui/atoms/image-item/image-item.vue
@@ -136,7 +136,7 @@
 @keyframes highlight-image {
 
     0% {
-        background-color: rgba(205, 255, 210, 0.3);
+        background-color: #aeffe2;
     }
 
     100% {
@@ -213,7 +213,7 @@ export const ImageItem = {
          */
         animationDuration: {
             type: Number,
-            default: 3000
+            default: 2000
         }
     },
     data: function() {
--- a/vue/components/ui/atoms/image-item/image-item.vue
+++ b/vue/components/ui/atoms/image-item/image-item.vue
@@ -38,12 +38,12 @@
         </div>
         <div
             class="item-image"
-            v-bind:style="style"
+            v-bind:style="imageStyle"
             v-bind:class="{ 'dropdown-open': optionsVisible }"
             ref="image"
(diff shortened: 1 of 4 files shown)


gcandal (approved):
@joamag apparently a "in your face" green actually works alright with the animation, even if in the static design it looks odd.

joamag:
@gcandal make the highlight color a property

BeeMargarida:
@gcandal I ended up going with the [Web animations API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API/Keyframe_Formats) because it was a solution that removed lots of code and made the feature more "concise". However, another approach might have the use of CSS variables (and maintaining the solution we had previously - css classes).

vue/components/ui/atoms/image-item/image-item.vue
```diff
@@ -40,9 +40,10 @@
             class="item-image"
             v-bind:style="style"
             v-bind:class="{ 'dropdown-open': optionsVisible }"
+            ref="image"
             v-on:animationend="onAnimationEnd"
         >
-            <image-ripe v-bind:style="imageStyle" v-bind:src="imageUrl" v-bind:alt="name" />
+            <image-ripe v-bind:style="style" v-bind:src="imageUrl" v-bind:alt="name" />
```
gcandal, line 46: Keep calling it `imageStyle` as we reserve `style` for when it's just for the root component (the div, in this case)

BeeMargarida: It will result in two computed that have the same code

gcandal (approved):

joamag (approved):

PR #472, opened 2021-03-31
fix: image-item highlight if true when mounted

BeeMargarida:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-twitch/issues/84#issuecomment-811251719<br>When the `highlight` prop was true when the component first mounted, the animation did not occur. This occured in `twitch-admin-ui` after going from white to admin-ui, where the first image-item was mounted with the highlight at `true` but the animation did not occur. |
| Dependencies | |
| Decisions | Play the animation when the component is first mounted (if `highlightData` is `true`). Done on mounted because it needs the `$refs.image` to be populated. Extracted animation logic to method. |
| Animated GIF | -- |

### DIFF
--- a/vue/components/ui/atoms/image-item/image-item.vue
+++ b/vue/components/ui/atoms/image-item/image-item.vue
@@ -245,8 +245,9 @@ export const ImageItem = {
         }
     },
     mounted: function() {
-        if (!this.highlightData) return;
-        this.animate();
+        // trigger the highlight animation
+        // if needed
+        if (this.highlightData) this.animate();
     },
     methods: {
         animate() {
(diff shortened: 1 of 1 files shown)


gcandal:
Use `immediate: true` in the watcher instead, as it allows you to keep a clean `created`.

gcandal:
> > Use `immediate: true` in the watcher instead, as it allows you to keep a clean `created`.
> 
> Can't because the `$refs` is still not populated

Oh you're using mounted and not created, nevermind.

gcandal (approved):

joamag (approved):

PR #484, opened 2021-04-29
feat: all option in select-checkboxes

gcandal:
| - | - |
| --- | --- |
| Issue | Motivated by an ad-hoc implementation in https://github.com/ripe-tech/ripe-util-vue/pull/53 |
| Animated GIF |  |


joamag:
I believe we should create some unit testing for this one ;)

joamag (approved):

PR #491, opened 2021-05-13
feat: added header prop

veryprofessionaldodo:
| - | - |
| --- | --- |
| Decisions | Added prop to control header visibility. Defaults to "true". If header is de-selected and the add button position is "top", it automatically sets it to the bottom. |
| Animated GIF | |

With the header:
image

Without the header:
image

### DIFF
--- a/vue/components/ui/organisms/input-list/input-list.vue
+++ b/vue/components/ui/organisms/input-list/input-list.vue
@@ -1,9 +1,10 @@
 <template>
     <div class="input-list">
-        <div class="list-header" v-if="header">
+        <div class="list-header" v-if="headerVisibility">
             <div
                 v-bind:class="`label ${field.value}`"
                 v-for="field in fieldLabels"
+                v-show="header"
                 v-bind:key="field.value"
             >
                 {{ field.label === undefined ? field.value : field.label }}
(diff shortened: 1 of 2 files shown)


3rdvision (approved):

vue/components/ui/organisms/input-list/input-list.vue
```diff
@@ -155,7 +159,7 @@ export const InputList = {
     computed: {
         buttonContainerStyle() {
             const base = {};
-            if (this.buttonAddRow === "top") {
+            if (this.buttonAddRow === "top" && this.header) {
```
gcandal, line 162: Use a new prop for this so that the position of the add button is not defined in terms of the header's visibility. Something like `addPosition` (top/bottom)

veryprofessionaldodo: But that is what the `buttonAddRow` prop already does. We just added the behaviour to automatically place it on the bottom if there's no header, otherwise there was an overlap between the button to add a new list, and the button to remove the first element of the list. 

Without that check, this is what happens when the `buttonAddRow` is set to `top`, and the `header` is de-selected:
image

veryprofessionaldodo: This has changed in the latest commits, now if the `header` prop is set to false and the `buttonAddRow` is set to `top`, the labels are removed, but the button is still in the same position. The header is only fully removed if the `header` prop is false, and the `buttonAddRow` is set to `bottom`.

gcandal (approved):

joamag:
@gcandal we need to fix this one
1. Make the plus button a simple static layout button
2. Make the header visibility a simple v-if operations

gcandal:
Closing in favor of https://github.com/ripe-tech/ripe-components-vue/pull/494

PR #501, opened 2021-05-27
Select: add input filter

3rdvision:
| - | - |
| --- | --- |
| Decisions | - Add `<input>` filter to `<select>` as now it makes sense to have this component considering all the use cases in which there are a lot of options. <br> - Based on https://github.com/ripe-tech/ripe-components-vue/pull/252 |
| Animated GIF | Peek 2021-05-27 12-46 |


vue/components/ui/molecules/select/select.stories.js
```diff
@@ -14,39 +14,50 @@ storiesOf("Components/Molecules/Select", module)
...
+                    { value: "option_5", label: "Daniel" },
+                    { value: "option_6", label: "Gabriel" },
+                    { value: "option_61", label: "Gabriel11" },
+                    { value: "option_62", label: "Gabriel12" },
+                    { value: "option_63", label: "Gabriel3" },
+                    { value: "option_7", label: "Geraldo" },
+                    { value: "option_8", label: "João" },
+                    { value: "option_81", label: "João 1" },
+                    { value: "option_82", label: "João 2" },
+                    { value: "option_83", label: "João 3" },
```
gcandal, line 30: Let's have some cool demo that: change the numbers to some random name.

vue/components/ui/molecules/select/select.vue
```diff
@@ -264,11 +317,16 @@ export const Select = {
...
+            if (this.filter) {
+                this.filterText = "";
+                this.$nextTick(() => this.$refs.input.focus());
+            }
         },
         closeDropdown() {
             if (!this.visibleData) return;
             this.dehighlight();
             this.visibleData = false;
+            if (this.filter) this.$nextTick(() => this.$refs.selectButton.focus());
```
gcandal, line 329: Comment these

vue/components/ui/molecules/select/select.vue
```diff
@@ -346,13 +410,14 @@ export const Select = {
             this.toggleDropdown();
         },
         onSelectButtonEnterKey() {
+            if (!this.filteredOptions[this.highlighted]) return;
```
gcandal, line 413: comment

vue/components/ui/molecules/select/select.vue
```diff
@@ -346,13 +410,14 @@ export const Select = {
...
         onSelectButtonEnterKey() {
+            if (!this.filteredOptions[this.highlighted]) return;
             this.toggleDropdown();
         },
         onSelectButtonSpaceKey() {
             this.toggleDropdown();
         },
         onKey(key) {
-            if (!key) return;
+            if (!key || this.filter) return;
```
gcandal, line 420: comment

vue/components/ui/molecules/select/select.vue
```diff
@@ -398,7 +463,11 @@ export const Select = {
                 return;
             }
 
-            this.setValue(this.options[this.highlighted].value);
+            if (!this.filteredOptions[this.highlighted]) {
```
gcandal, line 466: Why is this needed?

vue/components/ui/molecules/select/select.vue
```diff
@@ -436,12 +509,22 @@ export const Select = {
...
+            if (!this.filter || !this.filterText) return this.options;
+            return this.options.filter(option => this.filterFunction(option, this.filterText));
+        },
         selectedOption() {
             return this.options ? this.options[this.valueIndex] : null;
         },
         buttonText() {
             return this.selectedOption ? this.selectedOption.label : this.placeholder;
         },
+        dropdownOwners() {
```
gcandal, line 522: Dead code?

gcandal (commented):
Can you write unit tests?

vue/components/ui/molecules/select/select.vue
3rdvision: this is for the case in which the user presses enter but there's nothing highlighted in the list considering what's written in the select filter.

I personally found this to be the best way to handle the case - no action.

image

vue/components/ui/molecules/select/select.vue
```diff
@@ -19,9 +19,27 @@
             </option>
         </select>
         <div class="select-container" v-bind:style="style" v-else>
+            <input-ripe
+                class="select-input"
+                v-bind:value.sync="filterText"
+                v-show="filter && visibleData"
```
gcandal, line 25: Move this to a computed, and use `!stuff` in the `v-show` of the select-button below, so it's clear that it's either one or the other.

gcandal (commented):
Let's go over this in person as it has a ton of stuff in here.

vue/components/ui/molecules/select/select.vue
```diff
@@ -122,10 +150,19 @@
...
 }
 
+.select .select-container .select-input:focus,
 .select .select-container .select-button:focus {
     border-color: $aqcua-blue;
 }
 
+.select.select-filter .select-container .select-button {
+    transition: width 0.2s ease, background-color 0.2s ease;
+}
```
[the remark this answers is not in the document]
3rdvision: yes, to avoid a flicker in border color

vue/components/ui/molecules/select/select.vue
```diff
@@ -380,11 +447,11 @@ export const Select = {
         },
         onAltUpKey() {
             this.openDropdown();
-            this.highlight(0);
+            this.highlight(0, true);
```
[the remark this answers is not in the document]
3rdvision: yes, this 2 year bug still existed

vue/components/ui/molecules/select/select.vue
```diff
@@ -263,11 +313,19 @@ export const Select = {
...
             this.visibleData = true;
+
+            // focus input and with blank text after dropdown is opened
+            if (this.filter) {
+                this.filterText = "";
+                this.$nextTick(() => this.$refs.input.focus());
+            }
         },
-        closeDropdown() {
+        closeDropdown(blur = false) {
```
gcandal, line 323: Do it in the positive (focus = true)

vue/components/ui/molecules/select/select.vue
```diff
@@ -263,11 +313,19 @@ export const Select = {
...
+                this.$nextTick(() => this.$refs.input.focus());
+            }
         },
-        closeDropdown() {
+        closeDropdown(focus = true) {
             if (!this.visibleData) return;
             this.dehighlight();
             this.visibleData = false;
+            // focus the select button after dropdown is closed
+            if (this.filter && focus) this.$nextTick(() => this.$refs.selectButton.focus());
```
gcandal, line 328: It seems weird that I can ask `closeDropdown(focus = true)` and then no focus happens because I don't have a `flter`. Having this said, you can achieve the same result and keep this function sane by moving the `this.filter` logic to the caller of this function

vue/components/ui/molecules/select/select.vue
```diff
@@ -411,7 +483,12 @@ export const Select = {
                 return;
             }
 
-            this.setValue(this.options[this.highlighted].value);
+            // if there's no options for current filter, do nothing
+            if (!this.filteredOptions[this.highlighted]) {
```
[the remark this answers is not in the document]
gcandal: Bump. I believe you can reuse from 477-482 in `onEnterKey`

vue/components/ui/molecules/select/select.vue
3rdvision: Focus is already happening by default if I don't have a `filter`. This is to fix the behavior of the filter because it would focus an element that would hide right after.

But there indeed a problem with this because when I do `closeDropdown(focus = false)` it will in fact focus in non-filter mode.

3rdvision: refactoring..

vue/components/ui/molecules/select/select.vue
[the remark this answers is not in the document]
3rdvision: I remember we decided long ago to duplicated onEnterKey and onSpaceKey code [as it is in master ](https://github.com/ripe-tech/ripe-components-vue/blob/60e300cc47c5a7e1b6918c31ede756df0f7805e5/vue/components/ui/molecules/select/select.vue#L456)

simplifying this logic

vue/components/ui/molecules/select/select.vue
3rdvision: thinking better, there are a lot of places (8 places) that call closeDropdown and want `focus = false`
maybe it's better to just remove focus arg? as we will probably always want to focus the selectButton after closeDropdown

gcandal: If you can always call focus without any issue then yes, I'd do that.

3rdvision: I renamed to `maintainFocus` as removing the arg would stop global click from unfocusing.

Focus will always maintain itself in select for filter and non-filter mode.
The rename gives better info that if focus wasn't there in the beginning, it will not focus.

gcandal (approved):

gcandal:
@3rdvision don't forget the changelog entry 🚀

joamag:
Closing this until a proper rationale exists

PR #508, opened 2021-06-08
fix: console error call

gcandal:


joamag (approved):

gcandal:
Marked this as requiring backport but couldn't find any other uses of `alert`

PR #513, opened 2021-06-17
#99: Button group component

BeeMargarida:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-util-vue/issues/99 |
| Dependencies | -- |
| Decisions | • Button toggle component <br> • Button group toggle |
| Animated GIF | button-groupbutton-toggle|

### DIFF
--- a/vue/components/ui/molecules/button-toggle/button-toggle.vue
+++ b/vue/components/ui/molecules/button-toggle/button-toggle.vue
@@ -139,7 +139,6 @@ export const ButtonToggle = {
     },
     methods: {
         onClick(event) {
-            if (!this.toggle) return;
             this.$emit("click", event);
         }
     }
--- a/vue/components/ui/molecules/button-toggle/button-toggle.vue
+++ b/vue/components/ui/molecules/button-toggle/button-toggle.vue
@@ -81,6 +81,13 @@ export const ButtonToggle = {
             default: null
         },
         /**
+         * Whether it is able to be toggled or not.
+         */
+        toggle: {
+            type: Boolean,
+            default: true
+        },
+        /**
          * The value of the button.
          */
         value: {
@@ -132,6 +139,7 @@ export const ButtonToggle = {
     },
     methods: {
         onClick(event) {
+            if (!this.toggle) return;
             this.$emit("click", event);
         }
     }
--- a/vue/components/ui/molecules/button-toggle/button-toggle.vue
+++ b/vue/components/ui/molecules/button-toggle/button-toggle.vue
@@ -1,0 +1,150 @@
+<template>
+    <button-color
+        class="button-toggle"
+        v-bind:class="classes"
+        v-bind:text="textComputed"
+        v-bind:icon="iconComputed"
+        v-bind:color="colorComputed"
+        v-bind:disabled="disabled"
+        v-bind:min-width="0"
+        v-bind="buttonProps"
+        v-on:click="onClick"
+    />
+</template>
+
+<style lang="scss" scoped>
+.button-toggle.default {
+    border-radius: 5px;
+}
+
+.button-toggle.left {
+    border-radius: 5px 0px 0px 5px;
+}
+
+.button-toggle.middle {
+    border-radius: 0px;
+}
+
+.button-toggle.right {
+    border-radius: 0px 5px 5px 0px;
+}
+</style>
+
+<script>
+export const ButtonToggle = {
+    name: "button-toggle",
+    props: {
+        /**
+         * The name of the primary icon to appear in the middle of the button.
+         * This icon will show when `value` is true.
+         */
+        icon: {
+            type: String,
+            default: null
+        },
+        /**
+         * The name of the icon to appear in the middle of the button.
+         * This icon will show when `value` is false.
+         */
+        iconSecondary: {
+            type: String,
+            default: null
+        },
+        /**
+         * The color of the button when `value` is false.
+         */
+        color: {
+            type: String,
+            default: "white"
+        },
+        /**
+         * The color of the button when `value` is true.
+         */
+        colorSecondary: {
+            type: String,
+            default: "black"
+        },
+        /**
+         * The text of the button, will be shown if no
+         * icon is provided.
+         */
+        text: {
+            type: String,
+            default: null
+        },
+        /**
+         * The value of the button.
+         */
+        value: {
+            type: Boolean,
+            default: false
+        },
+        /**
+         * The orientation of the button that modefied its borders
+         * ("default", "left", "center" and "right").
+         */
+        orientation: {
+            type: String,
+            default: "default"
+        },
+        /**
+         * If the button is disabled.
+         */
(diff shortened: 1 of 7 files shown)


vue/components/ui/molecules/button-toggle/button-toggle.vue
```diff
@@ -0,0 +1,150 @@
...
+        },
+        /**
+         * The value of the button.
+         */
+        value: {
+            type: Boolean,
+            default: false
+        },
+        /**
+         * The orientation of the button that modefied its borders
```
3rdvision, line 83: typo

vue/components/ui/molecules/button-toggle/button-toggle.vue
```diff
@@ -0,0 +1,150 @@
...
+        disabled: {
+            type: Boolean,
+            default: false
+        },
+        /**
+         * Button color props.
+         */
+        buttonProps: {
+            type: Object,
+            default: () => {}
```
3rdvision, line 102:
```suggestion
            default: () => ({})

```

vue/components/ui/organisms/button-group/button-group.vue
```diff
@@ -0,0 +1,67 @@
...
+    watch: {
+        value(value) {
+            this.valueData = value;
+        },
+        valueData(value) {
+            this.$emit("update:value", value);
+        }
+    },
+    methods: {
+        calculateOrientation(index) {
```
3rdvision, line 55: nice one

3rdvision (commented):
https://user-images.githubusercontent.com/24736423/122436244-92332e00-cf90-11eb-9e25-94686f69e388.mov

Clean code, just needs a little improvement
I think when we toggle a button to false, it should set the value of `<button-group>` to null but it's not happening as we can see in the video above.

vue/components/ui/organisms/button-group/button-group.vue
```diff
@@ -0,0 +1,69 @@
...
+            if (index === 0) return "left";
+            if (index === this.items.length - 1) return "right";
+            return "middle";
+        },
+        onClick(event, item) {
+            this.$set(this, "valueData", item.value);
+            console.log("val", this.valueData);
+            // this.valueData = item.value;
```
3rdvision, line 63: remove debug code

3rdvision (approved):

gcandal (approved):

gcandal:
@BeeMargarida conflicts

gcandal (approved):

joamag:
@BeeMargarida We must handle buttons that are not select buttons (eg: regular buttons)

3rdvision (commented):
At this point it's confusing what's expected from button-toggle and button-group.

Clean code :+1: . I think there might be just a small problem when we have `toggle: false` for a button-toggle.

So I'm talking about when we set the prop `toggle: false` for a specific button-toggle of the button-group. I would expect, in that situation, for that button to behave like a button-color in which it just emits a click when it's clicked.

The problem is button-color does in fact emit the click event but then it is not passed to the parent component (button-toggle) or even the parent of the parent (button-group). See GIF below
Only button-color is emitting the event.

Peek 2021-08-04 18-31

So from my understanding we need to find a way to let the user of the component know when the user clicks a button-toggle with `toggle: false`, we have no way of knowing this if only the child of the child emits the event.

Also, please solve the changelog conficts :smile:

veryprofessionaldodo:
@3rdvision Nice catch about the toggle. So if I understood correctly, what you're saying is that if the `button-group` has `toggle` set to `false`, then the `button-toggle` elements inside should still emit the `click` event as normal, right? If that's the case, then the latest commit addresses that, and now the event is emitted, even if the group has that variable set to `false`. 

And nice catch on the conflicts, I tend to forget to pull the target branch, I'll keep that in mind next! :D

3rdvision:
@veryprofessionaldodo 

I was thinking that this could be solved by not updating the button-toggle value when prop `toggle` is `false`. I was only surprised to find that this was implemented in a way that you have to control the button-toggle state value from the outside, I was expecting it to emit an `update:value` after internal valueData change along with always emitting the `click` event.


@veryprofessionaldodo After your commit, `toggle` prop is dead code in toggle-button so I think the component might need a refactor to solve this.

@BeeMargarida do you remember why state change was done in the story instead of internally in the component just like [button-toggle-icon's implementation](https://github.com/ripe-tech/ripe-components-vue/blob/master/vue/components/ui/molecules/button-icon-toggle/button-icon-toggle.vue#L90)?
```
        methods: {
            onClick(event) {
                this.valueData = !this.valueData;
            }
        },
```

I could be missing something but looks like a good approach could be doing something similar to button-icon-toggle by creating a `valueData` and emitting `update:value` on change.

BeeMargarida:
> @veryprofessionaldodo
> 
> I was thinking that this could be solved by not updating the button-toggle value when prop `toggle` is `false`. I was only surprised to find that this was implemented in a way that you have to control the button-toggle state value from the outside, I was expecting it to emit an `update:value` after internal valueData change along with always emitting the `click` event.
> 
> @veryprofessionaldodo After your commit, `toggle` prop is dead code in button-toggle so I think the component might need a refactor to solve this.
> 
> @BeeMargarida do you remember why state change was done in the story instead of internally in the component just like [button-toggle-icon's implementation](https://github.com/ripe-tech/ripe-components-vue/blob/master/vue/components/ui/molecules/button-icon-toggle/button-icon-toggle.vue#L90)?
> 
> ```
>         methods: {
>             onClick(event) {
>                 this.valueData = !this.valueData;
>             }
>         },
> ```
> 
> I could be missing something but looks like a good approach could be doing something similar to button-icon-toggle by creating a `valueData` and emitting `update:value` on change.

Because button toggle does not save its state inside it, it is controlled by its parent, so that story had to control it.

3rdvision:
> Because button toggle does not save its state inside it, it is controlled by its parent, so that story had to control it. If I remember correctly

Makes sense.

@veryprofessionaldodo button-group just needs to emit the click
 
e.g. `items` prop:
```json
[
   {
      "value":"A",
      "label":"Item 1"
   },
   {
      "value":"B",
      "label":"Item 2"
   },
   {
      "value":"C",
      "label":"Item 3",
      "toggle": false
   }
]
```

When user clicks "Item 3" button, button-group should emit `click` with payload `C` and not update `valueData`.

veryprofessionaldodo:
@3rdvision , so if I'm understanding correctly, even if the button isn't toggled, the button group should still emit a `click` with the value of the selected item?

3rdvision:
> @3rdvision , so if I'm understanding correctly, even if the button isn't toggled, the button group should still emit a `click` with the value of the selected item?

Yes. The only difference being that if button item has `"toggle": true` it will emit the `update:value` and the `click` while a button that is not toggled

gcandal (approved):

joamag (approved):

PR #515, opened 2021-06-18
Error page: add navigate home button

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-util-vue/pull/124#issuecomment-863971011 |
| Decisions | - Add home page button. <br> - Fix home redirect for nuxt.js. |
| Animated GIF | image |

### DIFF
--- a/vue/components/ui/templates/error-part/error-part.vue
+++ b/vue/components/ui/templates/error-part/error-part.vue
@@ -11,7 +11,7 @@
             class="button-icon-home"
             v-bind:icon="homeIcon"
             v-bind:size="60"
-            v-if="navigation"
+            v-if="navigation && homeIcon"
             v-on:click="goHome"
         />
         <div class="error-message">
--- a/vue/components/ui/templates/error-part/error-part.vue
+++ b/vue/components/ui/templates/error-part/error-part.vue
@@ -1,11 +1,18 @@
 <template>
     <div class="error-part">
         <button-icon
-            class="button-icon-back"
+            class="button-icon back"
             v-bind:icon="'arrow-left'"
             v-bind:size="60"
             v-if="navigation"
             v-on:click="goBack"
+        />
+        <button-icon
+            class="button-icon home"
+            v-bind:icon="'home'"
+            v-bind:size="60"
+            v-if="navigation"
+            v-on:click="goHome"
         />
         <div class="error-message">
             <image-ripe
@@ -99,10 +106,17 @@
     margin: 24px auto 0px auto;
 }
 
-.button-icon-back {
-    left: 20px;
+.button-icon {
     position: absolute;
     top: 20px;
+}
+
+.button-icon.back {
+    left: 20px;
+}
+
+.button-icon.home {
+    left: 90px;
 }
 </style>
 
--- a/vue/components/ui/templates/error-part/error-part.vue
+++ b/vue/components/ui/templates/error-part/error-part.vue
@@ -116,7 +116,7 @@
 }
 
 .button-icon.home {
-    left: 90px;
+    right: 20px;
 }
 </style>
 
--- a/vue/components/ui/templates/error-part/error-part.vue
+++ b/vue/components/ui/templates/error-part/error-part.vue
@@ -143,6 +143,10 @@ export const ErrorPart = {
             type: String,
             default: null
         },
+        homeLocation: {
+            type: String,
+            default: "/"
+        },
         navigation: {
             type: Boolean,
             default: true
@@ -155,8 +159,7 @@ export const ErrorPart = {
     methods: {
         goHome() {
             if (!this.navigation) return;
-            if (window) window.location = "/";
-            this.$router.push({ name: "home" });
+            this.$router.push(this.homeLocation);
         },
         goBack() {
             if (!this.navigation) return;
--- a/vue/components/ui/templates/error-part/error-part.vue
+++ b/vue/components/ui/templates/error-part/error-part.vue
@@ -141,6 +141,7 @@ export const ErrorPart = {
             default: null
         },
         homeIcon: {
+            type: String,
             default: "home"
         },
         navigation: {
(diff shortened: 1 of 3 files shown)


vue/components/ui/templates/error-part/error-part.vue
```diff
@@ -138,6 +152,7 @@ export const ErrorPart = {
     methods: {
         goHome() {
             if (!this.navigation) return;
+            if (window) window.location = "/";
```
gcandal, line 155: use the router to push to "index"

vue/components/ui/templates/error-part/error-part.vue
```diff
@@ -99,11 +106,18 @@
...
     top: 20px;
 }
+
+.button-icon.back {
+    left: 20px;
+}
+
+.button-icon.home {
+    left: 90px;
+}
```
gcandal, line 120: Align to the right instead

vue/mixins/part.js
```diff
@@ -23,10 +24,16 @@ export const partMixin = {
...
+
+            const hasErrorPage = this.$root.$router.resolve({ name: "error" }).href !== "/";
+            if (hasErrorPage) {
+                this.$root.$router.push({
+                    name: "error",
+                    query: query
+                });
+            } else {
+                this.alertSimple(error.message);
+            }
```
gcandal, line 36: Keep the original implementation, let's force clients of this to provide an error page

vue/components/ui/templates/error-part/error-part.vue
```diff
@@ -1,12 +1,19 @@
...
     <div class="error-part">
         <button-icon
-            class="button-icon-back"
+            class="button-icon back"
             v-bind:icon="'arrow-left'"
             v-bind:size="60"
             v-if="navigation"
             v-on:click="goBack"
         />
+        <button-icon
```
gcandal, line 10: Let's go with allowing the specific logo (passed via prop) instead of using a generic icon.

Another possibility is trying to dynamically require `/res/logo.svg` and fallback to an icon if it doesn't exist (convention over configuration). Summing up:
0. Activate this button through a boolean prop
1. Try dynamic require
2. Use logo passed via prop
3. Fallback to this button icon

image

vue/components/ui/templates/error-part/error-part.vue
3rdvision: i tried this at first but it aparently it doesn't work in nuxt. I think it might be cause of the folder structure we have, it tries to pick a index/index.vue file and fails to find it.

we have no path or name for index so it looks as nothing will make router redirect to home, neither { path: "/" } works properly. This is why I opted for this pure js solution

gcandal: Are you saying that it's impossible to use the router to navigate to the index? That seems pretty odd

3rdvision: I also find it odd but none of these possibilities work:
```
this.$router.push({ name: "home" });
this.$router.push({ name: "index" });
this.$router.push({ path: "/" });
```

All of the above navigate to the not found path nuxt error page.

I think it might be the structure we're using.
Maybe we should be using path `/home/index.vue` for the home page and redirect the  "/" there.

vue/components/ui/templates/error-part/error-part.vue
3rdvision: Currently `<icon>` and `<button-icon>` only support string icon name as a prop.
A dynamic svg icon is possible by including the desired icon in project's own icon assets folder and pointing to it's name in the `icon` prop.
I see the benefit in requiring a dynamic icon as the choices might not satisfy the needs but not the benefit of requiring from paths other than the icon path such as `/res/logo.svg`.

Also, this seems looks far from pulse's header implementation:
image
image



I think for being able to dynamically require any asset from any other folder will require a components-vue refactor as currently it [only accounts for the icon name](https://github.com/ripe-tech/ripe-components-vue/blob/master/vue/components/ui/atoms/icon/icon.vue#L106).

I pushed a change to allow the use of custom icon which satisfies 2nd and 3rd point. The 1st point seems to require a lot of refactoring work for not so much gain. Is this what's intended or did I misunderstood something?

gcandal:
@3rdvision please ping me about this tomorrow

gcandal:
Try https://github.com/ripe-tech/ripe-twitch-ui/blob/b6091baab1b2f894b597624688d2b279da4a1efb/plugins/components-vue.js

gcandal:
Remove hover on header when header=False

vue/components/ui/templates/error-part/error-part.vue
```diff
@@ -1,12 +1,19 @@
 <template>
     <div class="error-part">
         <button-icon
-            class="button-icon-back"
```
gcandal, line 4: This is a breaking change and could be problematic because if automated tests depend on it, so keep it. You can add a button-icon class though.

vue/components/ui/templates/error-part/error-part.vue
```diff
@@ -126,6 +140,9 @@ export const ErrorPart = {
             type: String,
             default: null
         },
+        homeIcon: {
```
gcandal, line 143: add type

gcandal (approved):

vue/components/ui/templates/error-part/error-part.vue
```diff
@@ -126,6 +139,10 @@ export const ErrorPart = {
             type: String,
             default: null
         },
+        homeIcon: {
+            type: String,
+            default: "home"
```
joamag, line 144: make it null by default

vue/components/ui/templates/error-part/error-part.vue
joamag: let's try to find a better solution for this problem, `window.location` is not a very good solution

joamag (changes requested):

joamag (approved):

PR #518, opened 2021-07-05
version: 0.12.1

BeeMargarida:


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "ripe-components-vue",
-    "version": "0.12.0",
+    "version": "0.12.1",
```
gcandal, line 3: MAJOR.MINOR.PATCH:
- Major: breaking changes
- Minor: incremental change (new features)
- Patch: fixes

So this should be `0.13.0`

PR #524, opened 2021-07-22
#125 Select checkboxes can scroll to match when using input

veryprofessionaldodo:
| -| - | 
| ---| --- | 
| Issue | https://github.com/ripe-tech/ripe-util-vue/issues/125 |
| Dependencies | None |
| Decisions | Select checkboxes didn't have the feature to scroll to the closest match when input is given, so this PR adds that feature.  |
| Animated GIF | Peek 2021-07-22 09-31 |

### DIFF
--- a/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
+++ b/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
@@ -41,11 +41,11 @@
 </style>
 
 <script>
-import { utilsMixin } from "../../../../mixins";
+import { scrollMixin } from "../../../../mixins";
 
 export const SelectCheckboxes = {
     name: "select-checkboxes",
-    mixins: [utilsMixin],
+    mixins: [scrollMixin],
     props: {
         /**
          * Placeholder shown in the select button
@@ -181,13 +181,9 @@ export const SelectCheckboxes = {
             const index = this._items.findIndex(option =>
                 option.label?.toUpperCase().startsWith(keyBuffer)
             );
-
-            if (index === -1) return;
-
             const dropdown = this.$refs.select.$refs.dropdown.$refs.dropdown;
             const dropdownElement = dropdown.getElementsByClassName("dropdown-item")[0];
             const elements = dropdownElement.getElementsByClassName("checkbox-item");
-
             this.scrollToIndex(dropdown, elements, index);
         }
     }
--- a/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
+++ b/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
@@ -3,11 +3,17 @@
         class="select-checkboxes"
         v-bind:options="selectOptions"
         v-bind:value="'checkbox-group'"
+        v-bind:disabled="disabled"
+        v-bind="{
+            autoScroll: false,
+            maxHeight: 210,
+            ...selectProps
+        }"
         ref="select"
-        v-on:select:keydown="scrollToInput"
+        v-on:select:keydown="scrollToCheckbox"
     >
         <template v-slot:checkbox-group>
-            <div class="checkboxes" ref="checkboxes">
+            <div class="checkboxes" ref="checkboxes" v-on:click.stop>
                 <checkbox-group v-bind:items="_items" v-bind:values.sync="valuesData" />
             </div>
         </template>
@@ -171,7 +177,7 @@ export const SelectCheckboxes = {
         }
     },
     methods: {
-        scrollToInput(keyBuffer) {
+        scrollToCheckbox(_key, keyBuffer) {
             const index = this._items.findIndex(option =>
                 option.label?.toUpperCase().startsWith(keyBuffer)
             );
@@ -182,7 +188,7 @@ export const SelectCheckboxes = {
             const dropdownElement = dropdown.getElementsByClassName("dropdown-item")[0];
             const elements = dropdownElement.getElementsByClassName("checkbox-item");
 
-            this.scrollToIndexDropdown(dropdown, elements, index);
+            this.scrollToIndex(dropdown, elements, index);
         }
     }
 };
--- a/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
+++ b/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
@@ -177,7 +177,7 @@ export const SelectCheckboxes = {
         }
     },
     methods: {
-        scrollToCheckbox(_key, keyBuffer) {
+        scrollToCheckbox(key, keyBuffer) {
             const index = this._items.findIndex(option =>
                 option.label?.toUpperCase().startsWith(keyBuffer)
             );
--- a/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
+++ b/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
@@ -125,10 +125,10 @@ export const SelectCheckboxes = {
     },
     data: function() {
         return {
-            highlighted: null,
-            valuesData: this.values,
+            highlighted: null,            
             keyBuffer: "",
-            keyTimestamp: 0
+            keyTimestamp: 0,
+            valuesData: this.values
         };
     },
     computed: {
--- a/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
+++ b/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
@@ -1,18 +1,17 @@
 <template>
-    <div v-on:keydown.exact.stop="event => onKey(event.key)">
-        <select-ripe
-            class="select-checkboxes"
-            v-bind:options="selectOptions"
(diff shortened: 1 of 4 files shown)


vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -1,21 +1,18 @@
...
-            ...selectProps
-        }"
-    >
-        <template v-slot:checkbox-group>
-            <div class="checkboxes" v-on:click.stop>
-                <checkbox-group v-bind:items="_items" v-bind:values.sync="valuesData" />
-            </div>
-        </template>
-    </select-ripe>
+    <div v-on:keydown.exact.stop="event => onKey(event.key)">
```
NFSS10, line 2: This seems a very big change that has everything to break the select-checboxes in other places. I think this can't be done this way.

If the reason you added this div was just to do get the `on:key` event, try to just emit [this event in select-ripe](https://github.com/ripe-tech/ripe-components-vue/blob/a2b9d2aa0e199d59211232f6cf15d950ef052206/vue/components/ui/molecules/select/select.vue#L26),

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -170,6 +177,45 @@ export const SelectCheckboxes = {
             },
             deep: true
         }
+    },
+    methods: {
+        onKey(key) {
```
NFSS10, line 182: This seems to be [select-ripe `onKey` logic ](https://github.com/ripe-tech/ripe-components-vue/blob/a2b9d2aa0e199d59211232f6cf15d950ef052206/vue/components/ui/molecules/select/select.vue#L369) I would call the method something like `keyDebounce` and add this to a mixin (same for other logic you think it can be used in a general context), and share between the two components, or probably better yet, just have the select-ripe sending the key event and the `keyBuffer` with the `on:key` event and use that

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -170,6 +177,45 @@ export const SelectCheckboxes = {
...
+            this.keyBuffer += key.toUpperCase();
+            this.keyTimestamp = Date.now();
+            this.scrollToBestOption();
+        },
+        scrollToBestOption() {
+            const index = this._items.findIndex(option =>
+                option.label?.toUpperCase().startsWith(this.keyBuffer)
+            );
+
+            if (index === null || index < 0 || index >= this._items.length) return;
```
NFSS10, line 196: `if (index === -1) return;`

The return value is -1 if not found or >= 0, you shouldn't have all of this conditions, see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex#return_value

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -117,10 +114,20 @@ export const SelectCheckboxes = {
...
+         * Amount of time before current keybuffer is discarded.
+         */
+        keyTimeout: {
+            type: Number,
+            default: 500
         }
     },
     data: function() {
         return {
+            highlighted: null,
```
NFSS10, line 128: dead variable

NFSS10 (changes requested):

NFSS10:
@veryprofessionaldodo also, the build is failing

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
veryprofessionaldodo: Emitting an event in the select-ripe also worked, I did test that. I see what you mean with that, and I'll do those changes!

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
veryprofessionaldodo: Makes total sense, nice catch!

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
veryprofessionaldodo: That makes total sense, adding that in!

vue/components/ui/molecules/select/select.vue
```diff
@@ -373,6 +357,7 @@ export const Select = {
             }
             this.keyBuffer += key.toUpperCase();
             this.keyTimestamp = Date.now();
+            this.$emit("select:keydown", this.keyBuffer);
```
NFSS10, line 360: `this.$emit("select:keydown", key, this.keyBuffer);`

You should also emit the event and the key, As this `onKey` handler doesn't receive the `event` we are only emitting `key` :/

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -3,15 +3,11 @@
...
-        v-bind="{
-            autoScroll: false,
-            maxHeight: 210,
-            ...selectProps
-        }"
+        ref="select"
+        v-on:select:keydown="scrollToInput"
     >
         <template v-slot:checkbox-group>
-            <div class="checkboxes" v-on:click.stop>
```
NFSS10, line 14: Why did you remove this? This can break things in other places.

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -3,15 +3,11 @@
         class="select-checkboxes"
         v-bind:options="selectOptions"
         v-bind:value="'checkbox-group'"
-        v-bind:disabled="disabled"
-        v-bind="{
-            autoScroll: false,
-            maxHeight: 210,
-            ...selectProps
-        }"
```
NFSS10, line 11: Why did you remove this? This can break things in other places.

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -170,6 +169,21 @@ export const SelectCheckboxes = {
             },
             deep: true
         }
+    },
+    methods: {
+        scrollToInput(keyBuffer) {
```
NFSS10, line 174: `scrollToCheckbox`?

vue/mixins/utils.js
```diff
@@ -26,6 +26,24 @@ export const utilsMixin = {
...
+            }
+
+            const indexHeight = elements[index].offsetHeight;
+            const indexEnd = indexStart + indexHeight;
+
+            if (indexStart < visibleStart) {
+                dropdown.scrollTop = indexStart;
+            } else if (indexEnd > visibleEnd) {
+                dropdown.scrollTop = indexEnd - dropdown.clientHeight;
+            }
```
NFSS10, line 46: We can probably name this to be more generic. This should work with any tipe of container and not just with dropdowns right?

NFSS10: Also as we are making this a generic util method it's probably better to jsdoc it so other people can easily know what it does

NFSS10 (changes requested):

vue/mixins/utils.js
veryprofessionaldodo: Fair points, I'll do both!

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
veryprofessionaldodo: Yeah, I didn't like the name I had either, yours is more appropriate.

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
veryprofessionaldodo: It was meant to test things, as I was having difficulties getting the input. Forgot to put it back, similar to the `bind` and `disabled` parameters. Already re-added them, thanks!

NFSS10 (changes requested):

vue/mixins/utils.js
```diff
@@ -26,6 +26,31 @@ export const utilsMixin = {
...
         buildSlug(value) {
             return buildSlug(value);
+        },
+        /**
+         * Scrolls to the right element for a container,
+         * given the index of the element to scroll to.
+         * @param {*} container The parent container;
+         * @param {*} elements List of elements to iterate over;
+         * @param {*} index The index of the target element;
+         */
```
gcandal, line 36: Newline + types
```suggestion
        /**
         * Scrolls to the right element for a container,
         * given the index of the element to scroll to.
         *
         * @param {HTMLElement} container The parent container;
         * @param {HTMLCollection} elements List of elements to iterate over;
         * @param {Number} index The index of the target element;
         */

```

gcandal (approved):

joamag (approved):

joamag:
@veryprofessionaldodo good job... please confirm that the solution still works after the refactor

veryprofessionaldodo:
Can confirm, solution is working as expected using the latest commits from master! :)

PR #531, opened 2021-09-07
Upload Area: area container visibility prop

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-util-vue/issues/199 |
| Decisions | - Add `buttonText` to control the upload button text. <br> - Add `areContainer` prop to toggle the visibility of the upload-area-container extra area or hide it and just show the button.  |

https://user-images.githubusercontent.com/24736423/132326013-9a04e889-ffbd-46be-af7d-1e8828e84ca3.mov

PR #532, opened 2021-09-07
#199 Upload button: add new component and adapt to use it

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-util-vue/issues/199 |
| Decisions | - Split upload-area into [2 components as suggested](https://github.com/ripe-tech/ripe-components-vue/pull/531#issuecomment-914171003) <br> - Add upload-button <br> - Adapt upload-area <br> - Adapt chat and input-image |
| Animated GIF | image |

### DIFF
--- a/vue/components/ui/molecules/input-image/input-image.vue
+++ b/vue/components/ui/molecules/input-image/input-image.vue
@@ -1,7 +1,8 @@
 <template>
-    <upload-area
+    <upload-button
         class="input-image"
         v-bind:multiple="multiple"
+        v-bind:upload-button="false"
         v-bind:accept="accept"
         v-on:update:files="loadFiles"
     >
@@ -43,7 +44,7 @@
                 </div>
             </div>
         </template>
-    </upload-area>
+    </upload-button>
 </template>
 
 <style lang="scss" scoped>
@@ -56,12 +57,20 @@
     cursor: pointer;
 }
 
+.input-image.upload-button {
+    display: flex;
+}
+
 .input-image:hover {
     border-color: #1d1d1d;
 }
 
 .input-image.dragging {
     opacity: 0.3;
+}
+
+.input-image ::v-deep .button-upload {
+    display: none;
 }
 
 .input-image > .input-image-content {
--- a/vue/components/ui/molecules/input-image/input-image.vue
+++ b/vue/components/ui/molecules/input-image/input-image.vue
@@ -3,46 +3,45 @@
         class="input-image"
         v-bind:multiple="multiple"
         v-bind:accept="accept"
+        ref="uploadArea"
         v-on:update:files="loadFiles"
     >
-        <template v-slot="{ openModal }">
-            <div class="input-image-content" v-on:click="openModal">
-                <icon
-                    class="upload-icon"
-                    v-bind:icon="'add-camera'"
-                    v-bind:height="24"
-                    v-bind:width="24"
-                    v-show="!hasImages"
+        <div class="input-image-content" v-on:click="onInputImageContentClick">
+            <icon
+                class="upload-icon"
+                v-bind:icon="'add-camera'"
+                v-bind:height="24"
+                v-bind:width="24"
+                v-show="!hasImages"
+            />
+            <div
+                class="image-container"
+                v-for="(image, index) in imagesData"
+                v-bind:key="image.name"
+                v-on:click.stop
+            >
+                <lightbox
+                    class="image-preview"
+                    v-bind:image="image.data"
+                    v-bind:visible="visibleLightbox === index"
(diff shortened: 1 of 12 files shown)


vue/components/ui/organisms/chat/chat.vue
```diff
@@ -39,13 +39,17 @@
...
     box-sizing: border-box;
+    display: block;
+    height: initial;
     padding: 20px 20px 20px 20px;
     transition: opacity 0.125s ease-in;
+    white-space: initial;
 }
 
-.chat > .upload-area.dragging {
+.chat > div.upload-area.dragging {
```
gcandal, line 52: Are these changes really needed?

vue/components/ui/molecules/upload-button/upload-button.vue
```diff
@@ -0,0 +1,107 @@
+<template>
+    <div
+        class="upload-button"
```
gcandal, line 3: Not meant to make you miserable but I believe this is usually the other way around: `button-upload` (e.g.: `input-image`).

vue/components/ui/molecules/input-image/input-image.vue
```diff
@@ -3,57 +3,57 @@
         class="input-image"
         v-bind:multiple="multiple"
         v-bind:accept="accept"
+        ref="uploadArea"
         v-on:update:files="loadFiles"
     >
-        <template v-slot="{ openModal }">
```
gcandal, line 8: Can't you keep providing the `openModal` in the slot so that these changes aren't needed?

vue/components/ui/molecules/upload-area/upload-area.vue
```diff
@@ -2,45 +2,37 @@
...
-            </div>
+        <slot>
+            <transition name="fade-in" mode="out-in">
+                <div class="description" v-bind:key="dragging">
+                    {{ descriptionText }}
+                </div>
+            </transition>
+            <icon v-bind:icon="icon" v-bind:size="110" v-bind:color="'gray'" />
         </slot>
+        <upload-button
```
gcandal, line 18: Couldn't this simply be a substitute of the existing `button-color`, leaving the structure as it was? This change has some issues:
- Being outside the slot means it can't be overridden
- The former point plus the fact that the HTML changed is a breaking change of sorts

vue/components/ui/molecules/upload-area/upload-area.vue
```diff
@@ -121,90 +130,33 @@ export const UploadArea = {
...
-    watch: {
-        files(value) {
-            if (!value || value.length === 0) this.clear();
-            this.filesData = value;
-        },
-        dragging(value) {
-            this.$emit("update:dragging", value);
-        }
+    mounted: function() {
+        this.fileInputRef = this.$refs.uploadButton?.$refs?.filesInput;
```
gcandal, line 142: Create a `initUploadArea` method in the mixin that does this (setting the `fileInputRef` based on the provided argument), so that you encapsulate the logic there.

vue/components/ui/molecules/upload-area/upload-area.vue
```diff
@@ -121,90 +130,33 @@ export const UploadArea = {
...
-            return base;
         }
     },
     data: function() {
         return {
-            filesData: this.files,
             dragging: false,
-            draggingDisabled: false,
-            dragEnterTarget: null
+            fileInputRef: null
```
gcandal, line 138: Move to the mixin.

vue/components/ui/organisms/chat/chat.vue
3rdvision: yes, every line

vue/components/ui/molecules/upload-area/upload-area.vue
3rdvision: The problem is that previously we had this outside the slot:
```
        <input
            type="file"
            hidden
            v-bind:multiple="multiple"
            v-bind:accept="accept"
            ref="filesInput"
            v-on:change="onFilesInputChange"
        />
 ```
 Now that native input is inside `upload-button` we had to move it ouside the slot to replace the native input.
The native input has to somehow be outside the slot for the upload mixin logic to run using it.

An alternative could be moving back `<input>` in this component and toggling it between native or the one from upload-button when we have slots. I didn't do it because it went a little bit against the philosohpy of separating and re-using the components.

Visually, the upload-button becomes invisible with `display: none` when we the component receives slots so there's no visual impact in this change.

gcandal (approved):

joamag:
I'm merging this one but there's a small quick fix required after this.
In case the `multiple` prop is unset then the default label text should be "Upload File" singular and not plural.

PR #533, opened 2021-09-09
Listing: add key only option to addFilter

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-pulse/issues/234 |
| Decisions | - Add key only option to allow adding filters like addFilter("@today") resulting in `@today` instead of `@today=undefined` |

### DIFF
--- a/vue/components/ui/organisms/listing/listing.vue
+++ b/vue/components/ui/organisms/listing/listing.vue
@@ -401,9 +401,9 @@ export const Listing = {
         }
     },
     methods: {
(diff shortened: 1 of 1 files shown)


gcandal (approved):
👍 ! Changelog please.

joamag:
@gcandal this is a risky decision that may create regressions

joamag:
made the update less risky by avoiding empty string situations

PR #534, opened 2021-09-10
Button Upload: label and multiple fix

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-components-vue/pull/532#issuecomment-916710005 |
| Decisions | - Fix `buttonText` label when multiple is `true` <br> - Fix multiple files being set when dragging even when multiple was set to `false` |
| Animated GIF | Peek 2021-09-10 11-18 |

### DIFF
--- a/vue/components/ui/molecules/button-upload/button-upload.vue
+++ b/vue/components/ui/molecules/button-upload/button-upload.vue
@@ -18,7 +18,7 @@
         <slot v-bind:open-modal="openModal" />
         <button-color
             class="button-upload"
-            v-bind:text="buttonText"
+            v-bind:text="buttonText || buttonTextComputed"
             v-bind:icon="'cloud-upload'"
             v-bind:alignment="'center'"
             v-bind:disabled="disabled"
@@ -57,7 +57,7 @@ export const ButtonUpload = {
     props: {
         buttonText: {
             type: String,
-            default: "Upload Files"
+            default: null
         },
         disabled: {
             type: Boolean,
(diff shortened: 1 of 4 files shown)


vue/components/ui/molecules/button-upload/button-upload.vue
```diff
@@ -57,16 +57,12 @@ export const ButtonUpload = {
...
         buttonText: {
             type: String,
-            default: "Upload Files"
+            default: null
         },
         disabled: {
             type: Boolean,
             default: false
         },
-        multiple: {
```
gcandal, line 66: Why remove? You're using it below:

image

It seems to make sense to allow using the button to pick multiple files.

vue/mixins/upload.js
```diff
@@ -26,7 +30,7 @@ export const uploadMixin = {
             this.fileInputRef = filesInputRef;
         },
         setFiles(filesList) {
-            this.filesData = [...filesList];
+            this.filesData = this.multiple ? [...filesList] : [filesList[0]];
```
gcandal, line 33: Is this required? If you set `multiple="false"` in the `input` won't that guarantee you'll have a single one?

vue/components/ui/molecules/button-upload/button-upload.vue
3rdvision: This was moved to the mixin

vue/mixins/upload.js
3rdvision: This is to fix multiple files being set when dragging even when `multiple` is set to false.
This was an old bug. It was also happening in `upload-area` before the split component refactor.

gcandal (approved):

joamag (approved):

PR #536, opened 2021-09-23
ScrollMixin: Enhace scroll state reactivity

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-pulse/issues/240 |
| Decisions | - Change handlers and point them to `updateScrollState` that will maintain an internal current state of scroll size and positions. From there we compute the (reactive) values such as `shouldLoadMore`. <br> - Emit value changes so it's possible to access and sync important scrolling indicators such as `isScrollAtBottom` from outside the component.  |

### DIFF
--- a/vue/mixins/scroll.js
+++ b/vue/mixins/scroll.js
@@ -20,27 +20,54 @@ export const scrollMixin = {
     data: function() {
         return {
             scrollTop: false,
-            showScrollTop: false
+            scrollPositionY: 0,
+            scrollSizeY: 0
         };
     },
+    computed: {
+        isScrollAtTop() {
+            return this.scrollPositionY === 0;
+        },
+        isScrollAtBottom() {
+            return window.innerHeight + this.scrollPositionY >= this.scrollSizeY;
+        },
+        showScrollTop() {
+            return this.scrollPositionY > this.scrollTopThreshold;
+        },
+        shouldLoadMore() {
+            return this.scrollSizeY - this.loadingScrollOffset < window.innerHeight + this.scrollPositionY;
+        }
+    },
+    watch: {
+        isScrollAtTop(value) {
+            this.$emit("scroll-at-top", value);
+        },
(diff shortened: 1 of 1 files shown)


vue/mixins/scroll.js
```diff
@@ -20,27 +20,59 @@ export const scrollMixin = {
...
+        shouldLoadMore() {
+            return (
+                this.scrollSizeY - this.loadingScrollOffset <
+                window.innerHeight + this.scrollPositionY
+            );
+        }
+    },
+    watch: {
+        isScrollAtTop(value) {
+            this.$emit("scroll-at-top", value);
```
gcandal, line 46: remove emits

vue/mixins/scroll.js
```diff
@@ -20,27 +20,59 @@ export const scrollMixin = {
...
+        },
+        isScrollAtBottom(value) {
+            this.$emit("scroll-at-bottom", value);
+        },
+        shouldLoadMore(value) {
+            this.$emit("load-more", value);
+            if (value) {
+                this.getFilter().loadMore();
```
gcandal, line 54: only do this if getFilter is defined

vue/mixins/scroll.js
```diff
@@ -54,7 +79,8 @@ export const scrollMixin = {
         scrollToTop() {
             window.scroll({
                 top: 0,
-                left: 0
+                left: 0,
+                behavior: "smooth"
```
gcandal, line 83: This seems like a breaking change so we should avoid it, no?

gcandal, line 83: In any case, if this was needed it should be an option of the method.

3rdvision: I don't think it's a breaking change in the way that it could potentially break something for whoever's using.
If behavior smooth is not available, it will still scroll not smoothly

gcandal (approved):

PR #537, opened 2021-09-27
#240 Auto load more and forceItemsLoad

3rdvision:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-pulse/issues/240 |
| Dependencies | https://github.com/ripe-tech/ripe-components-vue/pull/536 |
| Decisions | - Add `forceItemsLoad` to listing and filter <br> - Add auto load after update items to listing |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * New `button-upload` component
 * Search clear prop on `listing`
 * Scrolling sync events like `isScrollAtBottom` on `listing`
+* Add `forceItemsLoad` to listing and filter
+* Add auto load after update items to listing
 
 ### Changed
 
--- a/vue/components/ui/molecules/filter/filter.vue
+++ b/vue/components/ui/molecules/filter/filter.vue
@@ -169,6 +169,10 @@ export const Filter = {
             type: Number,
             default: 200
         },
+        forceItemsToLoad: {
+            type: Boolean,
+            default: false
+        },
         useQuery: {
             type: Boolean,
             default: false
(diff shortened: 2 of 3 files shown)


gcandal:
@3rdvision please elaborate on why is this needed, as right now it's not very clear for me (and probably anyone else that checks this PR out)

3rdvision:
> @3rdvision please elaborate on why is this needed, as right now it's not very clear for me (and probably anyone else that checks this PR out)
@gcandal 

Explaining by feature:
## Add auto load after update items to listing

This could be considered a bug fix although nothing has been reported (yet). We can reproduce this when the [limit of listing](https://github.com/ripe-tech/ripe-components-vue/blob/master/vue/components/ui/organisms/listing/listing.vue#L313) is such that doesn't create a scroll in the page. For e.g if we define listing limit to be 5 this will happen (loadMore will stop even thought there are more orders):

Peek 2021-09-27 11-24

Notice that I had to zoom up in a way that I have a scroll to trigger the loadMore, otherwise it won't work

WIth this fix, this will be working as expected and even if we had a limit of 1, it would `loadMore` as I would expect it to:

Peek 2021-09-27 11-22

Note that I set up a `limit` of `1`, that's why the loading indicator is showing like above. If I changed to the default `limit: 25` there would be no difference visually or in the requests from what's already happening in `master`.

This fix is also needed for the issue because when we want to filter orders by past state, we need to fetch for states with the desired filter so later we can fetching for the orders of those states. In this scenario, there will be cases in which 25 states belong to a few orders instead of 25 new orders therefore when developing this feature I had a sittuation in which the first fetch was 2 listing items, the Listing component would think the fetching was over and loadMore would stop and that brings us to the second one.

## Add `forceItemsLoad` to listing and filter 

This feature will allow us to force the loadMore and control it from the outside by prop so we can decide from the outside if we should stop loadingMore or not.

This will all make more sense after Pulse's pull request which requires all these changes to work.

gcandal:
Closing for now as it lost relevance with the change of the approach to getting the last states.

PR #541, opened 2021-11-03
Listing: clickable rows

3rdvision:
| - | - |
| --- | --- |
| Issue | Currently the rows of the `<listing>` component are styled with `cursor: pointer` but actually not all that area is clickable in most cases.  |
| Decisions | - Add missing `clickableRows` prop so we're able to control table's prop from filter or listing component. <br> - Add missing `::v-deep .clickable` selector styling for specific items we want as clickable in listing |
| Animated GIF | Below |

### Before
Peek 2021-11-03 12-43

### After
Peek 2021-11-03 12-42

### DIFF
--- a/vue/components/ui/organisms/listing/listing.vue
+++ b/vue/components/ui/organisms/listing/listing.vue
@@ -160,6 +160,10 @@
     color: $link-hover-color;
 }
 
+::v-deep .clickable {
+    cursor: pointer;
+}
+
 .scroll-button {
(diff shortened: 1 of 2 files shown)


vue/components/ui/organisms/listing/listing.vue
```diff
@@ -159,6 +160,10 @@
     color: $link-hover-color;
 }
 
+::v-deep .clickable {
```
joamag, line 163: why is this needed?

joamag (changes requested):

PR #552, opened 2022-02-16
fix: apply correctly the border radius in `image-item`

BeeMargarida:
| - | - |
| --- | --- |
| Issue | The image did not have border radius applied:<br>image<br>Can be seen in `ripe-util-vue`'s storybook:<br>image |
| Dependencies | |
| Decisions | - Add `overflow: hidden` and `box-sizing: border-box` to `image-item` styling and reduce the image size by `1px` so that the border can be seen |
| Animated GIF | Now:<br>image<br>image|


BeeMargarida:
@joamag I checked with the previous version, now it has the same dimensions
image-item-comp

joamag (approved):

joamag:
@BeeMargarida I'm not sure if this requires back-port. Not sure if this component exists in any other library...

BeeMargarida:
> @BeeMargarida I'm not sure if this requires back-port. Not sure if this component exists in any other library...

Nope, only here

PR #561, opened 2022-03-15
#73 - Re-added support for tooltips in a Form Input

veryprofessionaldodo:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-white-admin-ui/issues/73 |
| Decisions | Support for `tooltip` in a `form-input` component, re-added due to changes in `tooltip` implementation. |
| Screenshots | image |

### DIFF
--- a/vue/components/ui/molecules/form-input/form-input.vue
+++ b/vue/components/ui/molecules/form-input/form-input.vue
@@ -1,15 +1,17 @@
 <template>
     <div class="form-input" v-bind:class="classes">
         <slot name="header">
-            <label-ripe
-                class="header"
-                v-bind:style="headerStyle"
-                v-bind:class="headerClasses"
-                v-bind:size="headerSize"
-                v-bind:text="header"
-                v-bind:for="id"
-                v-if="header"
-            />
+            <tooltip v-bind="{ ...tooltipProps }">
+                <label-ripe
+                    class="header"
+                    v-bind:style="headerStyle"
+                    v-bind:class="headerClasses"
+                    v-bind:size="headerSize"
+                    v-bind:text="header"
+                    v-bind:for="id"
+                    v-if="header"
+                />
+            </tooltip>
         </slot>
         <div class="flex-container">
             <div class="content">
@@ -51,6 +53,12 @@
 
 .form-input.form-input-inline .flex-container {
     flex: 1;
+}
+
+.form-input .header.tooltip {
+    text-decoration-line: underline;
+    text-decoration-style: dashed;
+    text-underline-offset: 3px;
 }
 
 .form-input .header,
@@ -148,6 +156,10 @@ export const FormInput = {
         footerMinWidth: {
             type: Number,
             default: null
+        },
+        tooltipProps: {
+            type: Object,
+            default: null
         }
     },
     computed: {
@@ -168,6 +180,7 @@ export const FormInput = {
         headerClasses() {
             const base = {};
             if (this.headerVariant) base[`${this.headerVariant}`] = true;
+            if (this.tooltipProps) base.tooltip = true;
             return base;
         },
         footerClasses() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-*
+* Add support for `tooltip` in `form-input` - [#73](https://github.com/ripe-tech/ripe-white-admin-ui/issues/73)
 
 ## [0.20.0] - 2022-03-02
 
--- a/vue/components/ui/organisms/form/form.vue
+++ b/vue/components/ui/organisms/form/form.vue
@@ -20,6 +20,7 @@
                             <form-input
                                 class="section-field"
                                 v-bind:classes="formInputClasses(field)"
+                                v-bind:tooltip-props="field.tooltipProps"
                                 v-bind:header="
                                     field.label !== undefined ? field.label : field.value
                                 "


vue/components/ui/molecules/form-input/form-input.vue
```diff
@@ -148,6 +156,10 @@ export const FormInput = {
         footerMinWidth: {
             type: Number,
             default: null
+        },
+        tooltipProps: {
+            type: Object,
+            default: null
```
NFSS10, line 162: Isn't this logging warnings? `Object` props needs a factory function. Reference: https://vuejs.org/guide/components/props.html#prop-validation

It should be:

```
tooltipProps: {
   type: Object,
   default: () => ({})
}
```

vue/components/ui/molecules/form-input/form-input.vue
```diff
@@ -168,6 +180,7 @@ export const FormInput = {
         headerClasses() {
             const base = {};
             if (this.headerVariant) base[`${this.headerVariant}`] = true;
+            if (this.tooltipProps) base.tooltip = true;
```
NFSS10, line 183: This shouldn't be needed, tooltip already has a class so if the idea is to have a class to select the header when using tooltip, you can do a selector like: `form-input .tooltip > .header`

vue/components/ui/molecules/form-input/form-input.vue
```diff
@@ -1,15 +1,17 @@
...
-            <label-ripe
-                class="header"
-                v-bind:style="headerStyle"
-                v-bind:class="headerClasses"
-                v-bind:size="headerSize"
-                v-bind:text="header"
-                v-bind:for="id"
-                v-if="header"
-            />
+            <tooltip v-bind="{ ...tooltipProps }">
```
NFSS10, line 4: No need for the spread operator, you can do `<tooltip v-bind="tooltipProps">`

vue/components/ui/molecules/form-input/form-input.vue
```diff
@@ -53,6 +55,12 @@
     flex: 1;
 }
 
+.form-input .header.tooltip {
```
NFSS10, line 58: Don't need the extra class, use this selector: `form-input .tooltip > .header`.

Also, isn't the color `#57626e` missing?

I think this is what the design wanted:
```
    text-decoration: underline #57626e dashed;
    text-underline-offset: 3px;
```

NFSS10 (changes requested):

vue/components/ui/molecules/form-input/form-input.vue
veryprofessionaldodo: No warnings were logged, but I wasn't aware of this! I'll change it up

vue/components/ui/molecules/form-input/form-input.vue
NFSS10: That's happening because of what we talked about now, tooltip can't be always in use or will break compatibility

vue/components/ui/molecules/form-input/form-input.vue
veryprofessionaldodo: Nice catch! Updated :)

vue/components/ui/molecules/form-input/form-input.vue
```diff
@@ -1,15 +1,10 @@
...
-                v-bind:class="headerClasses"
-                v-bind:size="headerSize"
-                v-bind:text="header"
-                v-bind:for="id"
-                v-if="header"
-            />
+            <tooltip v-bind="tooltipProps" v-if="hasTooltip">
+                <label-ripe class="header" v-bind="headerProps" v-if="header" />
+            </tooltip>
+            <label-ripe class="header" v-bind="headerProps" v-else />
```
NFSS10, line 7: This is changing the header logic, it should be `v-else-if="header"` instead of `v-else`

vue/components/ui/molecules/form-input/form-input.vue
```diff
@@ -53,6 +48,12 @@
     flex: 1;
 }
 
+.form-input .header.header-underline {
```
NFSS10, line 51: No need for this new `header-underlin` class. When `hasTooltip` is `tue`, the tooltip will be wrapping the header thus, you can do `.form-input .tooltip > .header`. Also remove that class `header-underline` from the computed `headerClasses` bellow

NFSS10 (changes requested):

vue/components/ui/molecules/form-input/form-input.vue
veryprofessionaldodo: You're right, this was a remnant of a previous version that remained, thanks!

vue/components/ui/molecules/form-input/form-input.vue
```diff
@@ -53,6 +48,12 @@
     flex: 1;
 }
 
+.form-input .tooltip .header {
```
NFSS10, line 51: Be as specific as you can, use the `>`:
`form-input .tooltip > .header`

NFSS10 (changes requested):

NFSS10 (approved):

vue/components/ui/molecules/form-input/form-input.vue
```diff
@@ -53,6 +48,12 @@
     flex: 1;
 }
 
+.form-input .tooltip > .header {
+    -webkit-text-decoration: underline #57626e dashed;
```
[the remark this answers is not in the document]
veryprofessionaldodo: Yup, otherwise Safari doesn't display any text decoration 🤷

vue/components/ui/molecules/form-input/form-input.vue
```diff
@@ -148,13 +149,29 @@ export const FormInput = {
...
     },
     computed: {
         headerStyle() {
             const base = { "min-width": `${this.headerMinWidth}px` };
             return base;
         },
+        hasTooltip() {
+            return Object.keys(this.tooltipProps).length !== 0;
+        },
+        headerProps() {
```
joamag, line 166: We typically don't use this pattern of assigning multiple props at the same type let's keep it explicitly as before.

vue/components/ui/molecules/form-input/form-input.vue
joamag, line 52: no need to use the `-webkit` prefixed values

joamag (changes requested):

vue/components/ui/molecules/form-input/form-input.vue
veryprofessionaldodo: @joaofsl Mentioned this as well, but in this case, if this property isn't here, Safari shows no text-decoration at all. Shouldn't this be preserved to maximise browser compatibility?

joamag (approved):

PR #562, opened 2022-03-15
fix: Immediately watch ´valueData´ in select component

beneditomauro:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-util-vue/issues/254 |
| Decisions | - Add `deep` flag to the `options` watcher <br> - Add `immediate` flag to the `options` watcher |


CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-*
+* Reset value on `options` change in `select` component
```
BeeMargarida, line 12: I know it is not related, but adding the issue that caused this is beneficial (perhaps add a comment in the Issue about the reasoning behind this change there)

vue/components/ui/molecules/select/select.vue
```diff
@@ -163,6 +163,10 @@ export const Select = {
             type: Array,
             default: () => []
         },
+        resetValueOnOptionsChange: {
```
BeeMargarida, line 166: that is a big name, perhaps it would be better to reduce it to something like `resetValue`, and jsdocs above it explaining what it does

NFSS10:
@beneditomauro @BeeMargarida Is this really needed?

As the options change comes from the parent, the parent should also be responsible for this behaviour so this shouldn't be a problem

CHANGELOG.md
NFSS10: @beneditomauro Add the issue link here, atleast the one who originated this problem

beneditomauro:
Yes, this was a be

> @beneditomauro @BeeMargarida Is this really needed?
> 
> As the options change comes from the parent, the parent should also be responsible for this behaviour so this shouldn't be a problem

FYI: Approach rethinked  after discussing with Nuno

CHANGELOG.md
```diff
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-*
+* Immediately emmit `update:value` event in `select` component. https://github.com/ripe-tech/ripe-util-vue/pull/219#issuecomment-1068134009
```
NFSS10, line 20: Format it like this:
```
[#number](link)
```

NFSS10 (changes requested):

NFSS10 (changes requested):
@beneditomauro Add bug label and uodate PR url to have the related issue number

NFSS10 (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+* Immediately emmit `update:value` event in `select` component - [#254](https://github.com/ripe-tech/ripe-util-vue/issues/254)
```
BeeMargarida, line 20:
```suggestion
* Immediately emit `update:value` event in `select` component - [ripe-util-vue/#254](https://github.com/ripe-tech/ripe-util-vue/issues/254)
```

BeeMargarida (commented):
Have you checked if current uses of `select` present throughout our codebase do not break with this change?

beneditomauro:
Ch

> Have you checked if current uses of `select` present throughout our codebase do not break with this change?


Does not seemed to break anything.
Checked✅:
- ripe-util-vue
- ripe-pulse
- ripe-sputnik
- ripe-orchestra

CHANGELOG.md
```diff
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-*
+* Immediately watch `valueData` event in `select` component - [#254](https://github.com/ripe-tech/ripe-util-vue/issues/254)
```
BeeMargarida, line 20:
```suggestion
* Immediately watch `valueData` event in `select` component - [ripe-util-vue/#254](https://github.com/ripe-tech/ripe-util-vue/issues/254)
```

BeeMargarida (approved):

PR #302, opened 2020-05-11
Extra icons support

NFSS10:
| - | - |
| --- | --- |
| Issue | Needed "options" icon to be in `icons` component to be able to better refactor `details-ripe` |
| Dependencies | -- |
| Decisions | • Support the addition of new icons in the "extra" folder <br>• Added "options" icon (obtained from `details-ripe` asset folder |
| Animated GIF | Components-icon_extra_options<br><br>**Its the same used by `details-ripe`:**<br>May-11-2020 14-31-32 |

PR #564, opened 2022-03-17
Better medium defaults for 'button-color'

veryprofessionaldodo:
| - | - |
| --- | --- |
| Issue | Related to [#302](https://github.com/ripe-tech/ripe-pulse/issues/302) |
| Decisions | Currently, the `button-color` component, doesn't adapt well to forcing the height property in the loader and icon. This was occurring in `ripe-pulse` where manual styling was needed. This PR adds better functioning defaults for the `medium` size for the button. |
| Screenshots | image image |


beneditomauro (approved):

vue/components/ui/atoms/button-color/button-color.vue
```diff
@@ -124,6 +124,10 @@
     border-color: #2d2d2d;
 }
 
+.button-color.button-color.button-color-medium {
```
NFSS10, line 127: This doesn't seem right, shouldn't it be `.button-color.button-color-medium`

NFSS10: Also this needs to go before before line 77, before `.button-color.button-color-small`

NFSS10 (changes requested):

NFSS10 (approved):

veryprofessionaldodo:
Pinging @joamag , just to not let this get forgotten

veryprofessionaldodo:
@hugo-gomes Why were reviews re-requested?

veryprofessionaldodo:
@hugo-gomes Why were reviews re-requested?

NFSS10 (approved):

beneditomauro (approved):

joamag:
@joao-conde check this out

vue/components/ui/atoms/button-color/button-color.vue
```diff
@@ -252,6 +261,12 @@
     width: 32px;
 }
 
+.button-color.button-color-medium ::v-deep .loader > div {
+    height: 28px;
+    margin-left: 2px;
+    width: 28px;
+}
```
joao-conde, line 268: Same for margin-left, none of the other variants have it.

joao-conde (commented):
@veryprofessionaldodo I believe most of this CSS should be on the place where you are using the button. Because by changing the component you change all other places that used it, which were working fine, right?

vue/components/ui/atoms/button-color/button-color.vue
```diff
@@ -73,6 +73,10 @@
     vertical-align: middle;
 }
 
+.button-color.button-color-medium {
+    max-height: 34px;
+}
```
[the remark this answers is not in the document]
veryprofessionaldodo: You're right, they have the `height` defined, changing that now.

vue/components/ui/atoms/button-color/button-color.vue
veryprofessionaldodo: It's something that could be addressed on the other side, removing it

joao-conde:
@veryprofessionaldodo CI failing probably due to: https://github.com/ripe-tech/peri-shipping/pull/137#issuecomment-1135772027

joao-conde:
@veryprofessionaldodo the problem with this PR is that you are changing a component with specific CSS for a specific usage of the button medium (in this case the export CSV button).

For example in RIPE Utils we use the button, and with your changes the UI is affected:

ship-master

ship-pr

Likewise, there are other places that this change affects.

Instead I propose you open a PR on RIPE Pulse with:
- the custom CSS you need to make it look like the design in the page where you use the button
- use the `"small"` variant as it looks to be a better fit I think (it aligns with the search bar on Pulse)
- while you are there, add a padding to the "Group Orders" button as well (that only shows when you tick one order)

image

PR #565, opened 2022-03-21
#317: Justification translate to human-readable format

joao-conde:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-pulse/issues/317 |
| Dependencies | -- |
| Decisions | Mixin to fetch translations for justification codes to human-readable format. |

image

image

@BeeMargarida  @3rdvision was a bit unsure, maybe this should be moved to:
- commons JS
- or create a new SDK method with core endpoint
- or just leave it as a mixin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-*
+* `justificationByCode` mixin method - [ripe-pulse/#317](https://github.com/ripe-tech/ripe-pulse/issues/317)
 
 ### Changed
 
--- a/vue/mixins/ripe.js
+++ b/vue/mixins/ripe.js
@@ -2,6 +2,39 @@ import { hasPermission, toTokensM } from "yonius";
 
 export const ripeMixin = {
     methods: {
+        async justificationByCode({
+            context = null,
+            code = null,
+            codeFull = null,
+            other = "other;"
+        } = {}) {
+            // if the other code is used we return with
+            // the ad-hoc message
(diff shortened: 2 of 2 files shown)


BeeMargarida (approved):
I'll aprove it, but I don't know if this is the best place (as you mentioned in the PR description). Is this going to be used in other places? This could be moved both to SDK or even Pulse directly, I think it is a bit out of context here without being used in more than 1 place

3rdvision:
@joao-conde I'm also not sure where this should be but it can be used by ripe-pulse and ripe-util-vue so it sure could be a ripe mixin. As for ripe-commons-js it doesn't seem to have ripe methods so maybe this could be the place

joao-conde:
Used by Pulse by now, but I guess even Robin could use it in the future no? In that case, probably SDK or RIPE Commons. 

Could also just be a new route in https://github.com/ripe-tech/ripe-core/blob/master/src/ripe_core/controllers/api/justification.py though I am not sure what to call it.

What do you think? @BeeMargarida  @3rdvision

BeeMargarida:
> Used by Pulse by now, but I guess even Robin could use it in the future no? In that case, probably SDK or RIPE Commons.
> 
> Could also just be a new route in https://github.com/ripe-tech/ripe-core/blob/master/src/ripe_core/controllers/api/justification.py though I am not sure what to call it.
> 
> What do you think? @BeeMargarida @3rdvision

Robin does not use `this.$ripe` call like that but instead uses context, so as is it would not work...only would work if moved to the SDK

joao-conde:
Closing because a Vue mixin approach will not work for clients not written in Vue, for RIPE Robin for example

PR #567, opened 2022-03-25
#257 `code-editor` component

NFSS10:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-util-vue/issues/257 |
| Dependencies | <del>https://github.com/ripe-tech/ripe-components-vue/pull/566</del> |
| Decisions | Create `code-editor` component which is a text editor with syntax highlight support |

https://user-images.githubusercontent.com/22588915/160178117-b0a37373-26e1-4dc8-bbe8-dc6e7c6a1718.mp4


vue/components/ui/molecules/code-editor/code-editor.vue
```diff
@@ -0,0 +1,209 @@
...
+        },
+        type: {
+            type: String,
+            default: "json"
+        },
+        validate: {
+            type: Function,
+            default: null
+        },
+        debounceTime: {
```
joaofsl, line 128: Leave a description of what this is.
Also, isn't the default value a bit too low?

joaofsl (changes requested):

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-*
+* Add variant `red` to tooltip - [ripe-util-vue/#257](https://github.com/ripe-tech/ripe-util-vue/issues/257)
+* Create `code-editor` component - [ripe-util-vue/#257](https://github.com/ripe-tech/ripe-util-vue/issues/257)
```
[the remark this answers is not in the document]
NFSS10: No, for example, check changelogs below these

vue/components/ui/molecules/code-editor/code-editor.vue
NFSS10: It's literally the a debounce time. The value is good, I tried different ones and this felt right. It's high enough to do what is supposed to do and short enough to be responsive.

vue/components/ui/molecules/code-editor/code-editor.stories.js
```diff
@@ -0,0 +1,31 @@
+import { storiesOf } from "@storybook/vue";
+import { withKnobs, boolean, select } from "@storybook/addon-knobs";
+
+storiesOf("Components/Molecules/Code Editor", module)
+    .addDecorator(withKnobs)
+    .add("Code Editor", () => ({
+        props: {
+            value: {
+                default: JSON.stringify({ example: "exemple text", "example number": 123 }, null, 4)
```
joaofsl, line 9: I know it's just a story but you may want to correct the typo.

joaofsl (changes requested):

joaofsl (approved):

veryprofessionaldodo:
I found two issues, although I think one is more easy to fix than the other. 

image

In the screenshot above, I noticed consistently that causing errors in the first non-space character meant that the error would appear in the previous line. Another nice to have, is that if the error is in the first few lines, the variant for the tooltip could be "down", so that the contents of the message would not overflow to outside the area. It normally might not cause an issue, but if it is something like the storybook, there's a chance it is impossible to see the error.

NFSS10:
> I found two issues, although I think one is more easy to fix than the other.
> 
> image
> 
> In the screenshot above, I noticed consistently that causing errors in the first non-space character meant that the error would appear in the previous line. Another nice to have, is that if the error is in the first few lines, the variant for the tooltip could be "down", so that the contents of the message would not overflow to outside the area. It normally might not cause an issue, but if it is something like the storybook, there's a chance it is impossible to see the error.

For the first one, it's not something you can control, its the linter who is showning you the error and dictating where it is. You are being bias because of your perseption of where the error should be, but what the parser is seeing is just that the required chars aren't in the correct place, it's correct either perception.

For the second, the problem is with the way storybook builds their stories so the fix would be for that container have the proper overflow rules. Or you can also redo the tooltip component from the ground up to be properly injected above the target component.

veryprofessionaldodo (approved):

PR #576, opened 2022-04-08
feat: highlight checked rows

joao-conde:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-pulse/issues/281  |
| Dependencies | -- |
| Decisions | -- |
| Animated GIF | |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-*
+* Background color for checked rows in table component - [ripe-pulse/#281](https://github.com/ripe-tech/ripe-pulse/issues/281)
 
 ### Changed
 
--- a/vue/components/ui/molecules/filter/filter.vue
+++ b/vue/components/ui/molecules/filter/filter.vue
@@ -10,6 +10,7 @@
                 v-bind:reverse="reverse"
                 v-bind:alignment="tableAlignment"
                 v-bind:variant="tableVariant"
+                v-bind:row-selection="rowSelection"
                 v-bind:clickable-rows="clickableRows"
                 v-bind:checkboxes="checkboxes"
                 v-bind:checked-items="checkedItems"
@@ -141,6 +142,10 @@ export const Filter = {
         tableVariant: {
             type: String,
             default: null
+        },
+        rowSelection: {
+            type: Boolean,
+            default: false
         },
         clickableRows: {
             type: Boolean,
--- a/vue/components/ui/molecules/table/table.vue
+++ b/vue/components/ui/molecules/table/table.vue
@@ -139,6 +139,10 @@
     background-color: $selected-dark-color;
 }
 
+.table tbody tr.checked {
+    background-color: $selected-dark-color;
+}
+
 .table th {
     color: $label-color;
     font-size: 11px;
@@ -519,6 +523,7 @@ export const Table = {
         },
         rowClasses(item) {
             const base = {
+                checked: this.checkedItemsData[item.id],
                 selected: this.isRowSelected(item.id),
                 clickable: item.clickable === undefined ? this.clickableRows : item.clickable,
                 hoverable: item.hoverable === undefined ? this.hoverableRows : item.hoverable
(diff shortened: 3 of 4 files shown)


vue/components/ui/molecules/table/table.vue
```diff
@@ -519,6 +523,7 @@ export const Table = {
         },
         rowClasses(item) {
             const base = {
+                checked: this.checkedItemsData[item.id],
```
BeeMargarida, line 526: With this, this behaviour will be enabled, so `rowSelection` is not used. Have you checked if the `selected` style and the logic of `selectedRowData` is actually being used here? Because it seems that logic was abandoned, since `selectedIdData` is not being used anywhere and `selectedRow` is not being passed by filter. Perhaps these can be recycled.
If not, we can add a new variable similar to `rowSelection` but for checked

BeeMargarida (commented):
Please add the new variables to the story of the components, so that this feature can be easily tested

vue/components/ui/molecules/table/table.vue
joao-conde: I think `rowSelection` is for one selected row only (so, the last selected/checked row). Visually they amount to the same, because the styles do not differ here, but they could (so having different classes may be a good thing)

joao-conde: Selected logic is working, it works as the row that was clicked, even if outside the checkbox. Also, it amounts to the last row clicked. So there can be many checked, but only one marked as selected.

joao-conde: I added the `rowSelection` to propagate to the table component, meaning before was not passed, meaning it is not used anywhere for sure. So perhaps we could remove this logic. What do you think?

BeeMargarida: I would not remove it, since the behaviour is already there. I would pass props to allow it, like the `rowSelection` and `selectedRow`. But similar to this behaviour, a new prop should be passed to allow checked selection style.

BeeMargarida: One thing I would check is if `selectedIdData` is really being used anywhere or if it can be removed

BeeMargarida:
@joao-conde Also, this new behaviour needs to be backported to the table in ripe white, since they are very similar and the selected behaviour and checkeboxes already exist there.

vue/components/ui/molecules/filter/filter.stories.js
```diff
@@ -24,6 +24,9 @@ storiesOf("Components/Molecules/Filter", module)
                     { value: "car", label: "Car" }
                 ]
             },
+            rowSelection: {
```
BeeMargarida, line 27: need to pass it to the filter below

vue/components/ui/organisms/listing/listing.stories.js
```diff
@@ -17,6 +17,9 @@ storiesOf("Components/Organisms/Listing", module)
                     { value: "device", label: "Device" }
                 ]
             },
+            rowSelection: {
```
BeeMargarida, line 20: need to pass it to the listing below

BeeMargarida (approved):

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-*
+* Background color for checked rows in table component - [ripe-pulse/#281](https://github.com/ripe-tech/ripe-pulse/issues/281)
```
3rdvision, line 13: Please mention we're passing `rowSelection` from listing

vue/components/ui/molecules/filter/filter.vue
```diff
@@ -142,6 +143,10 @@ export const Filter = {
             type: String,
             default: null
         },
+        rowSelection: {
```
3rdvision, line 146: Just wanted to say that "selectableRows" would be a better name, similar to `clickableRows`... but I noticed `table` uses "rowSelection" prop name for quite some time and would be a breaking change :disappointed:

joamag (approved):

PR #270, opened 2020-02-18
#260 Created "breadcrumbs" component

NFSS10:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-components-vue/issues/260 |
| Decisions | Created "breadcrumbs" component |
| Animated GIF | Breadcrumbs |

PR #579, opened 2022-04-12
#270 `select-checkboxes` search bar

NFSS10:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-util-vue/issues/270 |
| Dependencies | -- |
| Decisions | Add search bar to `select-checkboxes` |

https://user-images.githubusercontent.com/22588915/163013102-e0a73982-4db8-43f0-8f56-066099d51a94.mp4

### DIFF
--- a/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
+++ b/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
@@ -6,6 +6,7 @@
         v-bind:value="'checkbox-group'"
         v-bind:disabled="disabled"
         v-bind="{
+            highlightable: false,
             autoScroll: false,
             maxHeight: 210,
             ...selectProps
@@ -34,17 +35,6 @@
 
 <style lang="scss" scoped>
 @import "css/variables.scss";
-
-.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item.selected,
-.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item.highlighted,
-.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item.hover,
-.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item.active,
-.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item:not(.disabled):hover,
-.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item:not(.disabled):active,
-.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item:not(.disabled).selected,
-.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item:not(.disabled).highlighted {
-    background-color: $white;
-}
 
 .select-checkboxes .checkboxes {
     box-sizing: border-box;
--- a/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
+++ b/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
@@ -198,7 +198,9 @@ export const SelectCheckboxes = {
             return [{ label: this._label, value: "checkbox-group" }];
         },
         filteredItems() {
-            return this.searchValue ? this._filterCheckboxes(this._items, this.searchValue) : this._items;
+            return this.searchValue
+                ? this._filterCheckboxes(this._items, this.searchValue)
+                : this._items;
         },
         allSelected() {
             return Boolean(this.valuesData[this.allValue]);
--- a/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
+++ b/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
@@ -225,9 +225,9 @@ export const SelectCheckboxes = {
             if (this.filterCheckboxes) return this.filterCheckboxes;
             return (items, filter) => {
                 return items.filter(item => {
-                    const label = item.label.toLowerCase();
-                    const searchValue = this.searchValue.toLowerCase();
-                    return label.includes(searchValue);
+                    const itemValue = `${item.label || item.value}`.toLowerCase();
+                    filter = filter.toLowerCase();
+                    return itemValue.includes(filter);
                 });
             };
         },
--- a/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
+++ b/vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
@@ -26,7 +26,7 @@
                     v-bind:value.sync="searchValue"
                     v-if="search"
                 />
-                <checkbox-group v-bind:items="_items" v-bind:values.sync="valuesData" />
+                <checkbox-group v-bind:items="filteredItems" v-bind:values.sync="valuesData" />
             </div>
         </template>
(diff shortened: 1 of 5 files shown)


vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -152,6 +221,21 @@ export const SelectCheckboxes = {
                 }))
             ];
         },
+        _filterCheckboxes() {
+            if (this.filterCheckboxes) return this.filterCheckboxes;
+            return (items, filter) => {
+                return items.filter(item => {
+                    const itemValue = `${item.label || item.value}`.toLowerCase();
```
joaofsl, line 228:
```suggestion
                    const itemValue = (item.label || item.value).toLowerCase();
```
Don't think we need to convert this to a string

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -122,14 +163,42 @@ export const SelectCheckboxes = {
         allValue: {
             type: String,
             default: "$ALL"
+        },
+        /**
+         * Whether to show the search bar.
```
joaofsl, line 168: whether or not*

joaofsl (approved):

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
NFSS10: Value could be something that isn't a string, it's safer doing this way

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -14,19 +15,59 @@
...
 .select-checkboxes .checkboxes {
     box-sizing: border-box;
     margin: 5px 15px 5px 15px;
     width: 100%;
 }
 
+.select-checkboxes.search .checkboxes {
+    display: flex;
+    flex-direction: column;
+    margin: 0px 0px 0px 0px;
```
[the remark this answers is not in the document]
NFSS10: I think it's the contrary :sweat_smile:
imagem

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
joaofsl: Can it be a Number too? Because if it can only a String we wouldn't need. In case it is a null or undefined we shouldn't even be converting I believe.
But the `${}` surely "solves" all of it

joaofsl (approved):

veryprofessionaldodo (approved):

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -14,19 +15,59 @@
...
+
+.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item.selected,
+.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item.highlighted,
+.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item.hover,
+.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item.active,
+.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item:not(.disabled):hover,
+.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item:not(.disabled):active,
+.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item:not(.disabled).selected,
+.select-checkboxes ::v-deep .dropdown-container .dropdown > .dropdown-item:not(.disabled).highlighted {
+    background-color: $white;
```
joamag, line 46: let's instead create a prop to control the background color / design of the dropdown. ::v-deep should be avoided

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -122,14 +163,42 @@ export const SelectCheckboxes = {
...
+         * Whether or not to show the search bar.
+         */
+        search: {
+            type: Boolean,
+            default: false
+        },
+        /**
+         * Set of props passed on to search.
+         */
+        searchProps: {
```
joamag, line 177: move this to the bottom of props definition

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -122,14 +163,42 @@ export const SelectCheckboxes = {
...
     },
     data: function() {
         return {
-            valuesData: this.values
+            valuesData: this.values,
+            searchValue: null
         };
     },
     computed: {
+        classes() {
```
joamag, line 197: classes should be the last computed item before the "private" ones

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -152,6 +221,21 @@ export const SelectCheckboxes = {
                 }))
             ];
         },
+        _filterCheckboxes() {
```
joamag, line 224: private items should come last in the definition

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
```diff
@@ -152,6 +221,21 @@ export const SelectCheckboxes = {
...
+            return (items, filter) => {
+                return items.filter(item => {
+                    const itemValue = `${item.label || item.value}`.toLowerCase();
+                    filter = filter.toLowerCase();
+                    return itemValue.includes(filter);
+                });
+            };
+        },
+        filteredItems() {
+            return !this.search || !this.searchValue
```
joamag, line 235: no need for the `!this.search` validation

joamag: use positive operation `return this.searchValue ? this._filterCheckboxes(this._items, this.searchValue) : this._items;`

joamag (changes requested):

vue/components/ui/molecules/select-checkboxes/select-checkboxes.vue
NFSS10: This isn't to customize the dropdown background, this was to fix an style issue where the dropdown color wasn't right.

If you compare the normal `select` to this `select-checkboxes` you can see that the default `background-color` doesn't match.

What is happening is that `checkbox-group` is only one item of the `dropdown`.
As it's the only item, it's always highlighted/selected, thus its color it's always the color of a `dropdown` item that is highlighted/selected istead of the actual default `$white` `background-color` of the select.

The difference without the fix:
imagem
imagem

After the fix:
imagem

joamag: OK got it. If the issue is related to the highlight of elements, let's just create a prop that enables/disables highlighting of elements and disable such value for this specific situation. This way we can avoid `v-deep` which is critical in providing good encapsulated elements that interact with each other through a properly defined interface (props).

joamag (approved):

joamag:
@NFSS10 we may have to backport the highlightable code to other implementations of the dropdown/select

Issue #571, opened 2022-04-04
Create new component


dsafonso:
## Rationale

Create a new component that group's avatars that have max used to calculate the extra avatar. the extra should be the last element and be displayed

## Description

Create a new component that group's avatars that have max used to calculate the extra avatar. the extra should be the last element and be displayed like this:
<img width="191" alt="Screen Shot 2022-03-29 at 10 43 48" src="https://user-images.githubusercontent.com/20085/160584480-3811ee66-fa5a-444b-b473-d227ecb062bb.png">

## Estimation

| Discipline | Estimation |
| --- | --- |
| **UI/UX Design** | X days  |  
| **Software Development** | 2 days |
| **QA** | 0.5 days |

veryprofessionaldodo:
@hugo-gomes Regarding this issue, looking at the design, are all avatars meant to be slightly overlapping, or is it just one group of (for example) 3 avatars, with one of the circles being the `+5`? If all of them are meant to be overlapping, it probably will generate too much visual noise when 40 or so avatars are displayed at once.

dsafonso:
Test OK :+1:

PR #582, opened 2022-04-19
#571 - Add Avatar List Component

veryprofessionaldodo:
| - | - |
| --- | --- |
| Issue | #571 |
| Decisions | Create a new component that groups avatars, with a max value. If the number of avatars exceeds the maximum number of values, a new circle will show the amount of missing elements. |

Here's how it was supposed to look design and feature wise: 

https://user-images.githubusercontent.com/22790704/164028184-df71eb47-b3cf-43a6-b35c-5350b13ff81e.mov

Here's the current state of the PR:

https://user-images.githubusercontent.com/22790704/164028724-45fab0d9-e253-49a4-849a-6eaed31f7a99.mov

### DIFF
--- a/vue/components/ui/molecules/avatar-list/avatar-list.stories.js
+++ b/vue/components/ui/molecules/avatar-list/avatar-list.stories.js
@@ -1,5 +1,5 @@
 import { storiesOf } from "@storybook/vue";
-import { withKnobs, number, select } from "@storybook/addon-knobs";
+import { boolean, number, select, withKnobs } from "@storybook/addon-knobs";
 
 storiesOf("Components/Molecules/Avatar List", module)
     .addDecorator(withKnobs)
@@ -32,6 +32,9 @@ storiesOf("Components/Molecules/Avatar List", module)
                     },
                     "medium"
                 )
+            },
+            expandable: {
+                default: boolean("Expandable", true)
             }
         },
         template: `
@@ -39,6 +42,7 @@ storiesOf("Components/Molecules/Avatar List", module)
                 v-bind:avatar-list="list"
                 v-bind:max-elems="maxElems"
                 v-bind:size="size"
+                v-bind:expandable="expandable"
             />
         `
     }));
--- a/vue/components/ui/molecules/avatar-list/avatar-list.stories.js
+++ b/vue/components/ui/molecules/avatar-list/avatar-list.stories.js
@@ -1,0 +1,44 @@
+import { storiesOf } from "@storybook/vue";
+import { withKnobs, number, select } from "@storybook/addon-knobs";
+
+storiesOf("Components/Molecules/Avatar List", module)
+    .addDecorator(withKnobs)
+    .add("Avatar List", () => ({
+        props: {
+            maxElems: {
+                default: number("Max Elems", 3)
+            },
+            list: {
+                default: () => [
+                    { src: "http://i.pravatar.cc/300" },
+                    { src: "http://i.pravatar.cc/600" },
+                    { src: "http://i.pravatar.cc/700" },
+                    { src: "http://i.pravatar.cc/100" },
+                    { src: "http://i.pravatar.cc/200" },
+                    { src: "http://i.pravatar.cc/400" },
+                    { src: "http://i.pravatar.cc/500" }
+                ]
+            },
+            size: {
+                default: select(
+                    "Size",
+                    {
+                        "Very Tiny": "very-tiny",
+                        Tiny: "tiny",
+                        Small: "small",
+                        Medium: "medium",
+                        Large: "large",
+                        "Very Large": "very-large"
+                    },
+                    "medium"
+                )
+            }
+        },
+        template: `
+            <avatar-list
+                v-bind:avatar-list="list"
+                v-bind:max-elems="maxElems"
+                v-bind:size="size"
+            />
+        `
+    }));
--- a/vue/components/ui/molecules/avatar-list/avatar-list.stories.js
+++ b/vue/components/ui/molecules/avatar-list/avatar-list.stories.js
@@ -5,19 +5,11 @@ storiesOf("Components/Molecules/Avatar List", module)
     .addDecorator(withKnobs)
     .add("Avatar List", () => ({
         props: {
-            maxElems: {
-                default: number("Max Elems", 3)
+            expandable: {
+                default: boolean("Expandable", true)
             },
-            list: {
-                default: () => [
-                    { src: "http://i.pravatar.cc/300" },
-                    { src: "http://i.pravatar.cc/600" },
-                    { src: "http://i.pravatar.cc/700" },
-                    { src: "http://i.pravatar.cc/100" },
-                    { src: "http://i.pravatar.cc/200" },
-                    { src: "http://i.pravatar.cc/400" },
-                    { src: "http://i.pravatar.cc/500" }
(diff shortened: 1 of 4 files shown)


vue/components/ui/molecules/avatar-list/avatar-list.vue
```diff
@@ -0,0 +1,163 @@
...
+    <div class="avatar-list">
+        <div
+            class="avatar-container"
+            v-bind:class="classes"
+            v-bind:style="{ 'z-index': index }"
+            v-for="(avatar, index) in avatarList"
+            v-bind:key="index"
+        >
+            <div v-if="index == maxElems">
+                + {{ avatarList.length - maxElems }}
```
joaofsl, line 11: this should be moved to a prop

vue/components/ui/molecules/avatar-list/avatar-list.vue
```diff
@@ -0,0 +1,163 @@
...
+.avatar-list > .avatar-container > * {
+    border: 2px solid #ffffff;
+}
+
+.avatar-list > .avatar-container:not(:first-child) {
+    transition: margin 0.3s;
+}
+
+.avatar-list > .avatar-container:not(:first-child).very-tiny {
+    margin-left: -10px;
```
joaofsl, line 32: I believe our standard doesn't allow for negative margins

joaofsl (changes requested):

vue/components/ui/molecules/avatar-list/avatar-list.vue
veryprofessionaldodo: In this case, the negative margins is to force the overlap. There was another solution previously in place, using absolute and relative positions. Unfortunately, after that solution was implemented I was told that the avatars also needed to expand when hovered, and that presented problems. When using absolute positions, the parent component has no width and height, so the hover presented problems because when expanding the user could hover over "nothing", forcing the collapse.

This solution doesn't use absolute and relative positioning, so the only way to enforce overlap is through negative margins, that turn positive when hovered.

vue/components/ui/molecules/avatar-list/avatar-list.vue
veryprofessionaldodo: A prop, or a computed variable? The text here is dynamic, since it depends on how many elements the list has, and the given `maxElems` prop.

joaofsl: Sorry, I meant a computed

vue/components/ui/molecules/avatar-list/avatar-list.vue
joaofsl: I agree with you that it is a simpler solution. I'm not sure about how strict we are with this standard of if it even is one at all so let's ignore it for the time being at least

joaofsl (approved):
@veryprofessionaldodo I'll move this to the next reviewer, but make sure to resolve the conflicts

vue/components/ui/molecules/avatar-list/avatar-list.vue
```diff
@@ -0,0 +1,166 @@
+<template>
+    <div class="avatar-list">
+        <div
+            class="avatar-container"
+            v-bind:class="classes"
+            v-bind:style="{ 'z-index': index }"
+            v-for="(avatar, index) in avatarList"
+            v-bind:key="index"
+        >
+            <div v-if="index == maxElems">
```
NFSS10, line 10: Missing class name

vue/components/ui/molecules/avatar-list/avatar-list.vue
```diff
@@ -0,0 +1,166 @@
...
+        </div>
+    </div>
+</template>
+
+<style lang="scss" scoped>
+.avatar-list {
+    display: flex;
+}
+
+.avatar-list > .avatar-container > * {
```
NFSS10, line 23: Make the selector more specific, avoid using wildcards

vue/components/ui/molecules/avatar-list/avatar-list.vue
```diff
@@ -0,0 +1,166 @@
...
+
+.avatar-list:hover > .avatar-container.expandable:not(:first-child).large {
+    margin-left: 8px;
+}
+
+.avatar-list:hover > .avatar-container.expandable:not(:first-child).very-large {
+    margin-left: 10px;
+}
+
+.avatar-list > .avatar-container > div {
```
NFSS10, line 79: Make the selector more specific, avoid using general element tags

vue/components/ui/molecules/avatar-list/avatar-list.vue
```diff
@@ -0,0 +1,166 @@
...
+    align-content: center;
+    background: #e1e1e1;
+    border-radius: 50%;
+    color: #545454;
+    display: flex;
+    justify-content: center;
+}
+
+.avatar-list > .avatar-container.very-tiny > div {
```
NFSS10, line 88: Make the selector more specific, avoid using general element tags. It applies to the rest of selectors that have the ` > div`

vue/components/ui/molecules/avatar-list/avatar-list.vue
```diff
@@ -0,0 +1,166 @@
...
+            class="avatar-container"
+            v-bind:class="classes"
+            v-bind:style="{ 'z-index': index }"
+            v-for="(avatar, index) in avatarList"
+            v-bind:key="index"
+        >
+            <div v-if="index == maxElems">
+                {{ remainingElements }}
+            </div>
+            <avatar v-bind="{ size: size, ...avatar }" v-else-if="index < maxElems" />
```
NFSS10, line 13: Why the `v-else-if`?
Isn't `v-else` enough?

joaofsl: It's to not display when `index > maxElems`

NFSS10: Oh, I see now, It's doing the for loop to all elements and not limiting the for loop to maxElems.

I would purpose then to limit the for loop and avoid unecessary iterations. Make a computed that limits the array size, or make the for loop run only `maxElems` times to avoid unecessary iterations and simplify this condition

vue/components/ui/molecules/avatar-list/avatar-list.vue
```diff
@@ -0,0 +1,166 @@
+<template>
+    <div class="avatar-list">
+        <div
+            class="avatar-container"
+            v-bind:class="classes"
+            v-bind:style="{ 'z-index': index }"
```
NFSS10, line 6: Make this a computed variable

NFSS10 (changes requested):

NFSS10 (approved):

vue/components/ui/molecules/avatar-list/avatar-list.stories.js
```diff
@@ -0,0 +1,48 @@
...
+                        Tiny: "tiny",
+                        Small: "small",
+                        Medium: "medium",
+                        Large: "large",
+                        "Very Large": "very-large"
+                    },
+                    "medium"
+                )
+            },
+            expandable: {
```
joamag, line 36: use the same order of the props definition

joamag (changes requested):

PR #588, opened 2022-05-04
Add slots to `input-symbol` component

veryprofessionaldodo:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-util-vue/issues/259 |
| Decisions | There was the need to have an `input` component capable of having icons before and after, so in order to preserve backwards compatibility with existing `input` usages, slots were added to `input-symbol` to allow this possibility. |
| Screenshots | image |


beneditomauro (changes requested):
Well, if this is a input that can have icons left or right, this should be a molecule with `<input />` and `<icon />` and not an input that can have whatever on the left and on the right.

Basically, if this is supposed to be a input with icon, use `icon` directly in the component

veryprofessionaldodo:
@beneditomauro I see what you mean. What would you suggest, that perhaps the naming of the file be changed? I think the flexibility is good, since I'm already using this on another PR with a loader, for example.

beneditomauro (changes requested):
LGTM. Just add a quick example with slots in storybook

beneditomauro (approved):

vue/components/ui/molecules/input-symbol/input-symbol.stories.js
```diff
@@ -105,6 +105,27 @@ storiesOf("Components/Molecules/Input Symbol", module)
                     v-bind:symbol="symbol"
                     v-bind:symbol-position="symbolPosition" />
                 <p>Text: {{ valueData }}</p>
+                <h3>Using Slots</h3>
```
joaofsl, line 108: I know this is very picky, but let's avoid using h3

joaofsl (approved):

joamag (approved):

PR #594, opened 2022-05-20
Support node `16` , `17` and `18`

NFSS10:
| - | - |
| --- | --- |
| Issue | Node `16` is LTS and currently CI/CD isn't running for that version. `ripe-components-vue` is a very active project, thus this project should have an active maintenance (relevant: https://github.com/ripe-tech/ripe-welcome/pull/187#issue-1241883511) |
| Dependencies | -- |
| Decisions | Added node `16`, `17` and `18` to github CI/CD |

Reference: https://nodejs.org/en/about/releases/

**NOTE:** Github jobs are failling because of the known storybook issue https://github.com/ripe-tech/ripe-welcome/pull/187#issue-1241883511

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,7 +5,7 @@ jobs:
     name: Build
     strategy:
       matrix:
-        node-version: [12, 13, 14, 16]
+        node-version: [12, 13, 14, 16, 17, 18]
     runs-on: ubuntu-latest
     container: node:${{ matrix.node-version }}
     steps:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,7 +5,7 @@ jobs:
     name: Build
     strategy:
       matrix:
-        node-version: [12, 13, 14]
+        node-version: [12, 13, 14, 16]
     runs-on: ubuntu-latest
     container: node:${{ matrix.node-version }}
     steps:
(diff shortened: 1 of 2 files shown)


BeeMargarida:
@NFSS10 You can also add node17 and node 18

NFSS10:
> @NFSS10 You can also add node17 and node 18

The question is, do we want to support those? And do we want to support those, now?

BeeMargarida:
> > @NFSS10 You can also add node17 and node 18
> 
> The question is, do we want to support those? And do we want to support those, now?

17 is already being supported in custom-reports, simple-proxy, so that one can be added here. 18 is optional, but is will be the next lts, so I don't think it is a bad idea

NFSS10:
> > > @NFSS10 You can also add node17 and node 18
> > 
> > 
> > The question is, do we want to support those? And do we want to support those, now?
> 
> 17 is already being supported in custom-reports and simple-proxy, so that one can be added here. 18 is optional, but is will be the next lts, so I don't think it is a bad idea

Ok, sounds good to me 👍

BeeMargarida (approved):
Tested with node 17 and 18 locally (with the storybook fixes that will come in a later PR), all working :+1:

3rdvision (approved):

joamag (approved):

Issue #305, opened 2020-05-13
RIPE Pulse Mobile menu Bug


jorgefernandess:
## Description

RIPE Pulse Mobile app switcher is changing it's position when clicked.
The component's width doesn't fit in the menu.

## Repro Steps

* Open RIPE Pulse Mobile version
* Click in app switcher

## Screens

![2020-05-13 14 46 09](https://user-images.githubusercontent.com/57152582/81820516-9fc5eb00-9528-11ea-87b8-583e3e2a3a5b.gif)

gcandal:
@jorgefernandess when opening bugs, assign them to @joamag with the `Triage` label.

joamag:
@gcandal I know we've fixed this in the past. We need to find a better solution this time ;)

joamag:
@jorgefernandess congrats on your first bug opening!

gcandal:
@FlaviooLima give this one a go!

FlaviooLima:
@gcandal aye aye

PR #598, opened 2022-06-02
#305 Fix `code-editor` `value` emit event

NFSS10:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-util-vue/issues/305 |
| Dependencies | -- |
| Decisions | The `$emit` event should be `update:value` instead of `value` |


joamag (approved):

joamag:
@NFSS10 we're sure that we won't see any regression here right? ;)

NFSS10:
> @NFSS10 we're sure that we won't see any regression here right? ;)
Nope, I checked :)

PR #288, opened 2020-03-18
#27: New home page

gcandal:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-design/issues/27 |
| Decisions |   |
| Animated GIF | image |

PR #599, opened 2022-06-03
#288 button-dropdown disabled prop

joaofsl:
| - | - |
| --- | --- |
| Issue | Issue surged with https://github.com/ripe-tech/ripe-util-vue/issues/288 where the `button-dropdown` needed to be disabled|
| Decisions | Add `disabled` prop to `button-dropdown`|
| Animated GIF ||


NFSS10 (approved):
@joaofsl I would probably also do this in the `onDropdownItemClicked` method but looks good!

```javascript
 onDropdownItemClicked(event, item) {
    if (this.disabled) return;
    this.$emit("click:item", event, item);
}
```

joaofsl:
> @joaofsl I would probably also do this in the `onDropdownItemClicked` method but looks good!
> 
> ```js
>  onDropdownItemClicked(event, item) {
>     if (this.disabled) return;
>     this.$emit("click:item", event, item);
> }
> ```

I didn't do it there because technically it could never get there if disabled

joamag (approved):

joamag:
@joaofsl please check if back-porting is required say to Ripe White, etc.

joaofsl:
@joamag this component only exists here

PR #609, opened 2022-09-19
#288 - dropdown-button color prop

joaofsl:
| - | - |
| --- | --- |
| Issue | https://github.com/ripe-tech/ripe-util-vue/issues/288 |
| Decisions | Add `color` prop to `dropdown-button`|
|Animated GIF|

### DIFF
--- a/vue/components/ui/molecules/button-dropdown/button-dropdown.vue
+++ b/vue/components/ui/molecules/button-dropdown/button-dropdown.vue
@@ -57,7 +57,7 @@
 }
 
 .button-dropdown > .button {
-    background-color: $white;
+    background-color: #000000;
     border: 1px solid #e4e8f0;
     box-sizing: border-box;
     color: $pale-grey;
@@ -95,7 +95,7 @@
 .button-dropdown > .button:hover {
     background-color: $black;
     border-color: $black;
-    color: $white;
+    color: #000000;
 }
 
 .button-dropdown > .button:active {
@@ -140,7 +140,7 @@
 
 .button-dropdown.button-dropdown-black > .button {
     background-color: #2d2d2d;
-    color: $white;
+    color: #000000;
 }
 
 .button-dropdown.button-dropdown-black > .button:hover,
--- a/vue/components/ui/molecules/button-dropdown/button-dropdown.vue
+++ b/vue/components/ui/molecules/button-dropdown/button-dropdown.vue
@@ -1,15 +1,27 @@
 <template>
     <div class="button-dropdown" v-bind:class="classes">
-        <div class="button button-primary" v-on:click="onPrimaryClick">
-            <template v-if="primaryIcon">
-                <icon class="icon icon-front" v-bind:icon="primaryIcon" v-bind:color="'#000000'" />
-                <icon class="icon icon-back" v-bind:icon="primaryIcon" v-bind:color="'#ffffff'" />
-            </template>
+        <div
+            class="button button-primary"
+            v-on:click="onPrimaryClick"
+            v-on:mouseover="primaryActive = true"
+            v-on:mouseout="primaryActive = false"
+        >
+            <icon
+                class="icon"
+                v-bind:icon="primaryIcon"
+                v-bind:color="buttonColorPrimary"
+                v-if="primaryIcon"
+            />
             <span class="label">{{ label }} </span>
         </div>
-        <div class="button button-secondary" ref="button-secondary" v-on:click="onSecondaryClick">
-            <icon class="icon icon-front" v-bind:icon="secondaryIcon" v-bind:color="'#000000'" />
-            <icon class="icon icon-back" v-bind:icon="secondaryIcon" v-bind:color="'#ffffff'" />
+        <div
+            class="button button-secondary"
+            ref="button-secondary"
+            v-on:click="onSecondaryClick"
+            v-on:mouseover="secondaryActive = true"
+            v-on:mouseout="secondaryActive = false"
+        >
+            <icon class="icon" v-bind:icon="secondaryIcon" v-bind:color="buttonColorSecondary" />
         </div>
         <dropdown
             v-bind:items="items"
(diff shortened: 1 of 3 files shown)


vue/components/ui/molecules/button-dropdown/button-dropdown.vue
```diff
@@ -192,7 +208,20 @@ export const ButtonDropdown = {
...
+            if (this.color === "black" && !this.primaryActive) return "#ffffff";
+            if (this.color === "black" && this.primaryActive) return "#000000";
+            if (this.primaryActive) return "#ffffff";
+            return "#000000";
+        },
+        buttonColorSecondary() {
+            if (this.color === "black" && !this.secondaryActive) return "#ffffff";
+            if (this.color === "black" && this.secondaryActive) return "#000000";
+            if (this.secondaryActive) return "#ffffff";
+            return "#000000";
```
NFSS10, line 224: The conditions here a little to redundant. You can reduce the number of checks

NFSS10 (changes requested):

vue/components/ui/molecules/button-dropdown/button-dropdown.vue
joaofsl: I believe this is the most readable one, but feel free to suggest something else

vue/components/ui/molecules/button-dropdown/button-dropdown.vue
```diff
@@ -1,15 +1,27 @@
...
-            <icon class="icon icon-front" v-bind:icon="secondaryIcon" v-bind:color="'#000000'" />
-            <icon class="icon icon-back" v-bind:icon="secondaryIcon" v-bind:color="'#ffffff'" />
+        <div
+            class="button button-secondary"
+            ref="button-secondary"
+            v-on:click="onSecondaryClick"
+            v-on:mouseover="secondaryActive = true"
+            v-on:mouseout="secondaryActive = false"
+        >
+            <icon class="icon" v-bind:icon="secondaryIcon" v-bind:color="buttonColorSecondary" />
```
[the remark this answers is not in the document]
NFSS10: Being an open source project you can't really say that. I believe that it may not have issues with our internal stuff but that doesn't mean it won't break other stuff.

You should atleast maintain those `icon-front` and `icon-back` classes as they won't affect your implementation and would still maintain some backwards compatibility.

vue/components/ui/molecules/button-dropdown/button-dropdown.vue
NFSS10:
```js
if (primaryActive) return color === "black" ? "#000000" : "#ffffff";
return color === "black" ? "#ffffff" : "#000000";
```

And the `buttonColorSecondary` one is similar to this code block

vue/components/ui/molecules/button-dropdown/button-dropdown.vue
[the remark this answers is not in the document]
joaofsl: Changing classes inside the component will not break any retro compatibility if the component is being used correctly, thus making this point invalid.
However, I'll add it to the component so we can move on with the review 💪🏼

vue/components/ui/molecules/button-dropdown/button-dropdown.vue
joaofsl: In my opinion this one is less readable but more concise. I'll use it!

NFSS10 (approved):

vue/components/ui/molecules/button-dropdown/button-dropdown.vue
```diff
@@ -155,6 +166,10 @@ export const ButtonDropdown = {
             type: String,
             default: "medium"
         },
+        color: {
+            type: String,
+            default: "white"
+        },
```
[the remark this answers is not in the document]
joao-conde: The whole component works only for "white" and "black", so this shouldn't be any color I agree. Like if I typo "red" this wont work. Do we have enum types for props? If so, use "white" and "black" enum values or just go with the variant approach Nuno said.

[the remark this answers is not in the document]
joaofsl: I see the point you're making, as this isn't exactly an open field where you can use any value, but I just don't think `variant` is a name that fits well.
The user would still have to "guess" what options where available even if it was named `variant`. So at least by naming it `color` where narrowing down the change of the prop being missused.

And yeah I don't think we are using enums for props

joao-conde (commented):
@joaofsl can we do this approach but instead of only black and white support a "primaryColor" and "secondaryColor" props instead of only black and white?

NFSS10:
> @joaofsl can we do this approach but instead of only black and white support a "primaryColor" and "secondaryColor" props instead of only black and white?

I think also `colorVariant` can be a good possibility

joao-conde (approved):

joaofsl:
@NFSS10 @joao-conde please give it a quick check again. Added primary and secondary color options (see gif in the description)

vue/components/ui/molecules/button-dropdown/button-dropdown.vue
```diff
@@ -193,6 +193,26 @@ export const ButtonDropdown = {
...
+        buttonColorPrimary() {
+            return this.primaryActive ? this.colorSecondary : this.colorPrimary;
+        },
+        buttonColorSecondary() {
+            return this.secondaryActive ? this.colorSecondary : this.colorPrimary;
+        },
+        styleButtonPrimary() {
+            const base = {};
+            base.color = this.primaryActive ? this.colorSecondary : this.colorPrimary;
+            base["background-color"] = this.primaryActive ? this.colorPrimary : this.colorSecondary;
```
joao-conde, line 206: I would maintain the border-color property as well and would write, for consistency:

```js
base["color"] = ...
```

joao-conde: Check border color with design

joao-conde (changes requested):

vue/components/ui/molecules/button-dropdown/button-dropdown.vue
joaofsl: Design team agreed the best approach is to always have the border to be this light grey color

joao-conde (approved):

PR #614, opened 2022-10-27
feat: add reacticity to `emptySlots()` computed property in `molecules/upload-area.vue`

beneditomauro:
| - | - |
| --- | --- |
| Issue | - |
| Dependencies | - |
| Decisions |Conceptually, `computed` values in `vue` are used when logic needs to be triggered when component internal `state` changes weather the logic is conditional to the state itself. In this specific case, `emptySlots()` is running only one time on component creation, limiting use-cases when the parent component adds or removes slots accordingly (see screenshots below) not removing the button when a slot is added. What was done to fix this:  <br> <br> - Reference the respective data property to react to, and use it in a way that doesn't alter the returning boolean expression  |
| Animated GIF | No slots: Screenshot 2022-10-27 at 15 56 46 With slots:Screenshot 2022-10-27 at 15 57 12 |

### DIFF
--- a/vue/components/ui/molecules/upload-area/upload-area.vue
+++ b/vue/components/ui/molecules/upload-area/upload-area.vue
@@ -150,13 +150,10 @@ export const UploadArea = {
             return this.draggingIcon ? this.draggingIcon : "cloud-upload";
         },
         emptySlots() {
-            // forces this property to refresh on data change
-            const bindRefresh = Boolean(this.filesData) || !this.filesData;
-
             return (
+                !this.noFileSelected &&
                 Object.entries(this.$slots).length === 0 &&
-                !this.$scopedSlots?.default &&
-                bindRefresh
+                !this.$scopedSlots?.default
             );
         }
     }
--- a/vue/components/ui/molecules/upload-area/upload-area.vue
+++ b/vue/components/ui/molecules/upload-area/upload-area.vue
@@ -150,7 +150,14 @@ export const UploadArea = {
             return this.draggingIcon ? this.draggingIcon : "cloud-upload";
         },
         emptySlots() {
-            return Object.entries(this.$slots).length === 0 && !this.$scopedSlots?.default;
+            // forces this property to refresh on data change
+            const bindRefresh = Boolean(this.filesData) || !this.filesData;
+
+            return (
+                Object.entries(this.$slots).length === 0 &&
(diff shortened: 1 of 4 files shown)


3rdvision:
@beneditomauro are we able to reproduce this bug in components-vue storybook?

beneditomauro:
> @beneditomauro are we able to reproduce this bug in components-vue storybook?

Not as it is, we would have to add a story that inserts a slot after the component mounting

vue/components/ui/molecules/upload-area/upload-area.vue
```diff
@@ -150,7 +150,14 @@ export const UploadArea = {
...
         emptySlots() {
-            return Object.entries(this.$slots).length === 0 && !this.$scopedSlots?.default;
+            // forces this property to refresh on data change
+            const bindRefresh = Boolean(this.filesData) || !this.filesData;
+
+            return (
+                Object.entries(this.$slots).length === 0 &&
+                !this.$scopedSlots?.default &&
+                bindRefresh
+            );
```
3rdvision, line 160:
```suggestion
            return !this.noFileSelected && Object.entries(this.$slots).length === 0 && !this.$scopedSlots?.default;
```

use this instead and add this value to mixins/upload.js
```

    computed: {
        noFileSelected() {
            return this.filesData?.length === 0;
        }
    },
```

also, remove it from `input-file` component.

I think it looks more elegant. What do you say?

3rdvision (changes requested):

vue/mixins/upload.js
```diff
@@ -34,6 +34,11 @@ export const uploadMixin = {
             this.$emit("update:dragging", value);
         }
     },
+    computed: {
+        noFileSelected() {
```
3rdvision, line 38: nice!
Also, don't forget to remove this one from `input-file` component since it is already importing this mixin :smile:

3rdvision:
@beneditomauro tests are failing... can you check it out?

beneditomauro:
> @beneditomauro tests are failing... can you check it out?

This is not related to the PR, this is locally in master fails too:

Screenshot 2022-11-02 at 15 12 29

3rdvision (approved):
Approved and we need to solve the failing build in a separate PR

joamag (approved):

PR #627, opened 2023-01-26
refactor: cleanup dead code var in slot

3rdvision:
| - | - |
| --- | --- |
| Issue | Cleanup dead code variable `item` in slot that generates vue warns (only visible after linking ripe-pulse and ripe-components-vue for instance). Note that: in order to make a v-bind:var="var" inside a slot, this slot needs to be inside a template that receives this item. This was probably done by mistake from copy/paste of previous examples.
image|
| Decisions | - Cleanup dead code var that generates a spam of warnings |


3rdvision:
No changelog because it makes no sense to mention this kind of change.

joao-conde (approved):

PR #631, opened 2023-04-26
Only support Node.js LTS versions (14, 16, 18)

NFSS10:
| - | - |
| --- | --- |
| Issue | odd versions shouldn't be supported + node 12 should be deprecated as even node 14 will stop being supported at the end of this month |
| Dependencies | -- |
| Decisions | * Only support Node.js LTS versions (`14`, `16`, `18`) <br>* Update `deploy.yml` Node.js version to `18` |

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,7 +8,7 @@ jobs:
     name: Deploy
     strategy:
       matrix:
-        node-version: [12]
+        node-version: [18]
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


3rdvision (commented):
@NFSS10 would be cool to also change the [deployed](https://github.com/ripe-tech/ripe-components-vue/blob/master/.github/workflows/deploy.yml#L11) version to either 14, 16, or 18 to fit this update.

3rdvision (approved):

Issue #637, opened 2023-10-30
Create "Details Expandable" component


pmpc94:
Create "Details Expandable" component

Props should accept an array of arrays, e.g:

`[[ "name": "Pedro", "phone": 000000000, "email": pedro@platforme.com" ]]` where each entry corresponds to a row and each key to a column

It should be expandable:
![image](https://github.com/ripe-tech/ripe-components-vue/assets/30175756/b1b19f2a-cb80-486e-a8ea-439a026b0f97)

It should allow multiple data per row:
![image](https://github.com/ripe-tech/ripe-components-vue/assets/30175756/ed7a4f9c-2dec-4cea-b60a-38d4a76f3ca1)

PR #640, opened 2023-10-31
637 - create details expandable component

pmpc94:
### Issue
- https://github.com/ripe-tech/ripe-components-vue/issues/637

### Dependencies
- None

### Decisions
- Create a Details Expandable component that can be expandable
- Allow field customization according to any design by the use of slots

### Screenshots

https://github.com/ripe-tech/ripe-components-vue/assets/30175756/b10e353d-325c-4fe2-9832-cf1ffffc93be

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,7 @@
 module.exports = {
     extends: ["hive/vue", "hive/prettier"],
     rules: {
-        "vue/no-v-html": ["off"]
+        "vue/no-v-html": ["off"],
+        "linebreak-style": "off"
     }
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,6 @@
 module.exports = {
     extends: ["hive/vue", "hive/prettier"],
     rules: {
-        "vue/no-v-html": ["off"],
-        "linebreak-style": "off"
+        "vue/no-v-html": ["off"]
     }
 };
--- a/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
+++ b/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
@@ -45,6 +45,17 @@ storiesOf("Components/Organisms/Details Expandable", module)
             <template v-slot="slotProps">
                 <input-ripe v-bind:value=slotProps.fieldValue></input-ripe>
             </template>
+            <template v-slot:customer-address>
+                <input-ripe v-bind:style="'width: 50%'" v-bind:value="data.customer.address.city"></input-ripe>
+            </template>
+            <template v-slot:shipper-address>
+                <input-ripe v-bind:style="'width: 50%'" v-bind:value="data.shipper.address.city"></input-ripe>
+            </template>
             </details-expandable-ripe>
         `
     }));
+
+
+
+
+
--- a/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
+++ b/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
@@ -54,14 +54,18 @@ storiesOf("Components/Organisms/Details Expandable", module)
                 v-on:discard:click="onDiscardClick"
                 v-on:save:click="onSaveClick"
             >
-            <template v-slot:first_row>Customized Customer</template>
-            <template v-slot:first_row-label-field_name>name</template>
-            <template v-slot:first_row-field_object>
-                <input-ripe v-bind:style="'width: 50%'" v-bind:value="data.first_row.field_object.property_a"></input-ripe>
-            </template>
-            <template v-slot:second_row-field_object>
-                <input-ripe v-bind:style="'width: 50%'" v-bind:value="data.second_row.field_object.property_c"></input-ripe>
-            </template>
+                <template v-slot:first_row>
+                    Customized Customer
+                </template>
+                <template v-slot:first_row-label-field_name>
+                    name
+                </template>
+                <template v-slot:first_row-field_object>
+                    <input-ripe v-bind:style="'width: 50%'" v-bind:value="data.first_row.field_object.property_a" />
+                </template>
+                <template v-slot:second_row-field_object>
+                    <input-ripe v-bind:style="'width: 50%'" v-bind:value="data.second_row.field_object.property_c" />
+                </template>
             </details-expandable-ripe>
         `
     }));
--- a/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
+++ b/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
@@ -1,0 +1,55 @@
+import { storiesOf } from "@storybook/vue";
+import { withKnobs } from "@storybook/addon-knobs";
+
+storiesOf("Components/Organisms/Details Expandable", module)
+    .addDecorator(withKnobs)
+    .add("Details Expandable", () => ({
+        props: {
+            data: {
+                type: Object,
+                default: () => ({
+                    customer: {
+                        name: "John Doe",
+                        phone: "351999999999",
+                        email: "john_doe@platforme.com",
+                        address: {
+                            line: "2540 Shore Blvd Apt LF",
+                            city: "Astoria",
+                            postal_code: "11102-3954",
+                            country_code: "US",
+                            state_code: "NY"
+                        }
+                    },
+                    shipper: {
+                        name: "Coolstuffz Porto",
+                        phone: "222333444",
+                        address: {
+                            line: "Rua de Santa Catarina 4",
+                            city: "Porto",
+                            postal_code: "4000542",
+                            country_code: "PT"
+                        }
+                    }
+                })
+            }
+        },
+        template: `
+            <details-expandable-ripe
+                v-bind:data="data"
+            >
+            <template v-slot="slotProps">
+            <label-ripe v-bind:text='slotProps.fieldName'></label-ripe>
+                <input-ripe v-bind:value=slotProps.fieldValue></input-ripe>
+            </template>
+            </details-expandable-ripe>
+        `
+    }));
+
+
+
+
+
+
+
+
+
--- a/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
+++ b/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
@@ -41,6 +41,7 @@ storiesOf("Components/Organisms/Details Expandable", module)
             <details-expandable-ripe
                 v-bind:data="data"
                 v-bind:label-font-size="fontSize"
+                v-bind:style="'margin-top: 25px'"
             >
             <template v-slot:customer-address>
                 <input-ripe v-bind:style="'width: 50%'" v-bind:value="data.customer.address.city"></input-ripe>
--- a/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
+++ b/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
@@ -8,26 +8,20 @@ storiesOf("Components/Organisms/Details Expandable", module)
             data: {
                 type: Object,
                 default: () => ({
-                    customer: {
-                        name: "John Doe",
-                        phone: "351999999999",
-                        email: "john_doe@platforme.com",
-                        address: {
-                            line: "2540 Shore Blvd Apt LF",
-                            city: "Astoria",
-                            postal_code: "11102-3954",
-                            country_code: "US",
-                            state_code: "NY"
+                    first_row: {
+                        field_a: "something",
+                        field_b: "something else",
+                        field_object: {
+                            property_a: "property a",
+                            property_b: "property b"
                         }
                     },
-                    shipper: {
-                        name: "Coolstuffz Porto",
-                        phone: "222333444",
-                        address: {
-                            line: "Rua de Santa Catarina 4",
-                            city: "Porto",
-                            postal_code: "4000542",
-                            country_code: "PT"
+                    second_row: {
+                        field_c: "yet something",
+                        field_d: "yet something else",
+                        field_object: {
+                            property_a: "property c",
+                            property_b: "property d"
                         }
                     }
                 })
--- a/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
+++ b/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
@@ -31,19 +31,24 @@ storiesOf("Components/Organisms/Details Expandable", module)
                         }
                     }
                 })
+            },
+            fontSize: {
+                type: Number,
+                default: 14
             }
         },
         template: `
             <details-expandable-ripe
                 v-bind:data="data"
+                v-bind:label-font-size="fontSize"
             >
             <template v-slot="slotProps">
-            <label-ripe v-bind:text='slotProps.fieldName'></label-ripe>
                 <input-ripe v-bind:value=slotProps.fieldValue></input-ripe>
             </template>
             </details-expandable-ripe>
         `
     }));
+
 
 
 
--- a/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
+++ b/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
@@ -48,13 +48,3 @@ storiesOf("Components/Organisms/Details Expandable", module)
             </details-expandable-ripe>
         `
     }));
-
-
-
-
-
-
-
-
-
-
--- a/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
+++ b/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
@@ -62,7 +62,3 @@ storiesOf("Components/Organisms/Details Expandable", module)
             </details-expandable-ripe>
         `
     }));
-
-
-
-
--- a/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
+++ b/vue/components/ui/organisms/details-expandable/details-expandable.stories.js
@@ -20,8 +20,8 @@ storiesOf("Components/Organisms/Details Expandable", module)
                         field_c: "yet something",
                         field_d: "yet something else",
                         field_object: {
-                            property_a: "property c",
-                            property_b: "property d"
+                            property_c: "property c",
+                            property_d: "property d"
                         }
                     }
                 })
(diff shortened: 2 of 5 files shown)


.eslintrc.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
     extends: ["hive/vue", "hive/prettier"],
     rules: {
-        "vue/no-v-html": ["off"]
+        "vue/no-v-html": ["off"],
+        "linebreak-style": "off"
```
joao-conde, line 5: Let's not change this config, it standardizes our line breaks.

vue/components/ui/organisms/details-expandable/details-expandable.stories.js
```diff
@@ -0,0 +1,64 @@
+import { storiesOf } from "@storybook/vue";
+import { withKnobs } from "@storybook/addon-knobs";
+
+storiesOf("Components/Organisms/Details Expandable", module)
+    .addDecorator(withKnobs)
+    .add("Details Expandable", () => ({
+        props: {
+            data: {
+                type: Object,
+                default: () => ({
```
joao-conde, line 10: Use some more generic default like the one from: https://github.com/ripe-tech/ripe-components-vue/blob/v-pca/637-create-details-expandable-component/vue/components/ui/organisms/container-form/container-form.stories.js#L15

Of course you can expand that if need be (pun intended)

vue/components/ui/organisms/details-expandable/details-expandable.stories.js
```diff
@@ -0,0 +1,64 @@
...
+                        address: {
+                            line: "Rua de Santa Catarina 4",
+                            city: "Porto",
+                            postal_code: "4000542",
+                            country_code: "PT"
+                        }
+                    }
+                })
+            },
+            fontSize: {
```
joao-conde, line 35: I dont see this prop in the component

vue/components/ui/organisms/details-expandable/details-expandable.stories.js
```diff
@@ -0,0 +1,64 @@
+import { storiesOf } from "@storybook/vue";
+import { withKnobs } from "@storybook/addon-knobs";
+
+storiesOf("Components/Organisms/Details Expandable", module)
+    .addDecorator(withKnobs)
+    .add("Details Expandable", () => ({
```
joao-conde, line 6: Some component props are missing from the story I think ?

vue/components/ui/organisms/details-expandable/details-expandable.vue
```diff
@@ -0,0 +1,156 @@
...
+        saveButtonText: {
+            type: String,
+            default: "Save changes"
+        }
+    },
+    data: function() {
+        const initialIsExpanded = {};
+        for (const sectionName in this.data) {
+            initialIsExpanded[sectionName] = false;
+        }
```
joao-conde, line 133: prefer `Object.keys()` over `for..in` making it clearer it is an object iteration over its keys and avoiding future mistakes:

```js
const isExpanded = {};
Object.keys(this.data).forEach(k => isExpanded[k] = false);
```

vue/components/ui/organisms/details-expandable/details-expandable.vue
```diff
@@ -0,0 +1,156 @@
...
+            type: String,
+            default: "Discard changes"
+        },
+        saveButtonText: {
+            type: String,
+            default: "Save changes"
+        }
+    },
+    data: function() {
+        const initialIsExpanded = {};
```
joao-conde, line 130: Would probably just call it isExpanded object

vue/components/ui/organisms/details-expandable/details-expandable.vue
```diff
@@ -0,0 +1,156 @@
+<template>
+    <div class="details-expandable">
+        <div v-bind:class="isExpanded[sectionName] ? 'details-expandable-section-show' : 'details-expandable-section-hide'" v-for="(section, sectionName, index) in data" v-bind:key="index">
+            <div class="details-expandable-section" v-on:click="toggleSection(sectionName)">
+                <h2 class="details-expandable-header">{{ sectionName }}</h2>
+                <icon class="details-expandable-caret" v-bind:icon="isExpanded[sectionName] ? 'chevron-up' : 'chevron-down'" />
+            </div>
+            <div class="details-expandable-row" v-for="(value, name, subIndex) in section" v-bind:key="subIndex">
+                <label-ripe class="details-expandable-title" v-bind:text="capitalizeName(name)" v-bind:font-size="labelFontSize" />
```
joao-conde, line 9: Maybe we should not capitalize the name in the component and instead let the component user capitalize the name if they want to (or apply any other transformation)

vue/components/ui/organisms/details-expandable/details-expandable.vue
```diff
@@ -0,0 +1,156 @@
...
+            </div>
+        </div>
+        <div class="details-expandable-btn-container">
+            <button-color
+                class="details-expandable-btn-discard"
+                v-bind:size="'small'"
+                v-bind:text="discardButtonText"
+                v-bind:secondary="true"
+                v-bind:icon="'close'"
+                v-on:click="event => onDiscardHandler(event)"
```
joao-conde, line 24: Can we just pass i the function like

```
v-on:click="onDiscardHandler"
```

joao-conde (changes requested):

vue/components/ui/organisms/details-expandable/details-expandable.vue
pmpc94: here I would at least use the name capitalization as the default as this names serves as a kind of title:
image

vue/components/ui/organisms/details-expandable/details-expandable.stories.js
pmpc94: added the rest

.eslintrc.js
pmpc94: for some reason I was always getting this error when this rule was enabled by default:

`expected linebreaks to be 'CRLF' but found 'LF'.` and this has something to do with the OS's (afaik), which can be different and bring this error more often

vue/components/ui/organisms/details-expandable/details-expandable.stories.js
pmpc94: isn't the one we have default enough?

vue/components/ui/organisms/details-expandable/details-expandable.vue
pmpc94: I created a new slot where the user is free to capitalize it how it wants. If he doesn't use it, then it will capitalize the first letter by default

.eslintrc.js
joao-conde: Yes, Linux and MacOS use LF line endings and Windows uses CRLF.

Simple run "npm run pretty" before commiting.

vue/components/ui/organisms/details-expandable/details-expandable.stories.js
joao-conde: It is shipping specific, when this component won't be specifically for shipping.

vue/components/ui/organisms/details-expandable/details-expandable.vue
joao-conde: Ok, seems a good approach

vue/components/ui/organisms/details-expandable/details-expandable.stories.js
```diff
@@ -0,0 +1,67 @@
...
+                v-on:discard:click="onDiscardClick"
+                v-on:save:click="onSaveClick"
+            >
+            <template v-slot:customer>Customized Customer</template>
+            <template v-slot:customer-label-name>name</template>
+            <template v-slot:customer-address>
+                <input-ripe v-bind:style="'width: 50%'" v-bind:value="data.customer.address.city"></input-ripe>
+            </template>
+            <template v-slot:shipper-address>
+                <input-ripe v-bind:style="'width: 50%'" v-bind:value="data.shipper.address.city"></input-ripe>
```
joao-conde, line 63: data.customer does not exist right

joao-conde (changes requested):

vue/components/ui/organisms/details-expandable/details-expandable.stories.js
pmpc94: sorry added you too soon, just changed it

joao-conde (approved):

joao-conde:
Not very familiar with components and I would like @3rdvision to take a look.

vue/components/ui/organisms/details-expandable/details-expandable.stories.js
```diff
@@ -0,0 +1,67 @@
...
+            >
+            <template v-slot:first_row>Customized Customer</template>
+            <template v-slot:first_row-label-field_name>name</template>
+            <template v-slot:first_row-field_object>
+                <input-ripe v-bind:style="'width: 50%'" v-bind:value="data.first_row.field_object.property_a"></input-ripe>
+            </template>
+            <template v-slot:second_row-field_object>
+                <input-ripe v-bind:style="'width: 50%'" v-bind:value="data.second_row.field_object.property_c"></input-ripe>
+            </template>
+            </details-expandable-ripe>
```
3rdvision, line 65: maybe pretty didn't pick this but we always make sure everything is well formatted, even when inside `template`
```suggestion
            <details-expandable-ripe
                v-bind:data="data"
                v-bind:label-font-size="labelFontSize"
                v-bind:style="'margin-top: 25px'"
                v-bind:discard-button-text="discardButtonText"
                v-bind:save-button-text="saveButtonText"
                v-on:discard:click="onDiscardClick"
                v-on:save:click="onSaveClick"
            >
                <template v-slot:first_row>
                    Customized Customer
                </template>
                <template v-slot:first_row-label-field_name>
                    name
                </template>
                <template v-slot:first_row-field_object>
                    <input-ripe
                        v-bind:style="'width: 50%'"
                        v-bind:value="data.first_row.field_object.property_a"
                    />
                </template>
                <template v-slot:second_row-field_object>
                    <input-ripe
                        v-bind:style="'width: 50%'"
                        v-bind:value="data.second_row.field_object.property_c"
                    />
                </template>
            </details-expandable-ripe>
```

3rdvision (commented):
Thanks for contributing to ripe-components-vue, even after vue3 deprecation and this mostly collecting dust after new products, there's still a lot of work history just sitting around here :disappointed: 

It might be frustrating to go full solo dive into old conventions and quality code standards especially when adding an organism component.

Have you considered extending/reusing the [form component](https://ripe-components-vue-now.platforme.com/?path=/story/components-organisms-form--form) and [section expandable](https://ripe-components-vue-now.platforme.com/?path=/story/components-molecules-section-expandable--section-expandable) instead of creating this one? We try to avoid duplicated component code as much as possible.

@NFSS10 can you check this one out since you were the one that wrote most of `form.vue` code?

pmpc94:
@3rdvision very good points, perhaps I should have pinned someone before I started to discuss some points 😅 

regarding your questions, yes I have considered reusing the `Form` and `Section` components but it would still require some changes because they are not really the same comparing with the new design. To avoid modifying these components I decided to create a new one but we could try to reuse them if you think is the best idea.

vue/components/ui/organisms/details-expandable/details-expandable.vue
```diff
@@ -0,0 +1,178 @@
...
+    margin-top: 10px;
+}
+
+.details-expandable-row {
+    margin-top: 25px;
+}
+</style>
+<script>
+export const DetailsExpandable = {
+    name: "details-expandable-ripe",
```
NFSS10, line 132: I think you can rename it `details-expandable`, no need for the ripe suffix. The `ripe` suffix was used in legacy components or in component that had conflicting names with standard HTML elements

vue/components/ui/organisms/details-expandable/details-expandable.vue
```diff
@@ -0,0 +1,178 @@
...
+    props: {
+        data: {
+            type: Object,
+            required: true
+        },
```
NFSS10, line 137: This type of prop is problematic and we should avoid it as it's too general. Usually we use Arrays (which isn't necessarily a lot more "specific") with a defined template on how to define key -> value pairs.

Also `data` is a very general name for a prop, if you can use a more informative name I think it would be best. Something like `fields` or something similar. Check and follow the standard in other components.

vue/components/ui/organisms/details-expandable/details-expandable.vue
```diff
@@ -0,0 +1,178 @@
...
+            type: String,
+            default: "Save changes"
+        }
+    },
+    data: function() {
+        const isExpanded = {};
+        Object.keys(this.data).forEach(k => {
+            isExpanded[k] = false;
+        });
+
```
NFSS10, line 156: Avoid computing data like this. Use other lifecycle's methods like `mounted/created` for that. Also use watches to handle these type of changes if necessary.

One of the reasons we do this is to make operations as general and reusable as possible

vue/components/ui/organisms/details-expandable/details-expandable.vue
```diff
@@ -0,0 +1,178 @@
...
+        return {
+            isExpanded
```
NFSS10, line 158: Follow the standard of other components, be a little more verbose when declaring data fields.
```suggestion
            isExpanded: isExpanded

```

vue/components/ui/organisms/details-expandable/details-expandable.vue
```diff
@@ -0,0 +1,178 @@
...
+        };
+    },
+    methods: {
+        capitalizeName(name) {
+            return name.charAt(0).toUpperCase() + name.slice(1);
+        },
```
NFSS10, line 164: I think we have a method for this already. I think you can use this:
https://github.com/ripe-tech/ripe-components-vue/blob/acbb133f34194cdff348f95484b819d72413e793/vue/components/ui/molecules/code-editor/code-editor.vue#L113

vue/components/ui/organisms/details-expandable/details-expandable.stories.js
```diff
@@ -0,0 +1,71 @@
...
+                            property_a: "property a",
+                            property_b: "property b"
+                        }
+                    },
+                    second_row: {
+                        field_c: "yet something",
+                        field_d: "yet something else",
+                        field_object: {
+                            property_c: "property c",
+                            property_d: "property d"
```
NFSS10, line 24: Avoid snake case in JS, use camel case,

vue/components/ui/organisms/details-expandable/details-expandable.stories.js
```diff
@@ -0,0 +1,71 @@
...
+                </template>
+                <template v-slot:first_row-label-field_name>
+                    name
+                </template>
+                <template v-slot:first_row-field_object>
+                    <input-ripe v-bind:style="'width: 50%'" v-bind:value="data.first_row.field_object.property_a" />
+                </template>
+                <template v-slot:second_row-field_object>
+                    <input-ripe v-bind:style="'width: 50%'" v-bind:value="data.second_row.field_object.property_c" />
+                </template>
```
NFSS10, line 68: Don't directly pass text to a slot, wrap it in a div.

If you want to only change the values use a prop instead. For example, if you only want to change the text in the header, instead of passing text via a slot, create a prop that allows to change the header text.

If it's a field that is dynamically generated, make the prop that allows to generate that dynamic structure accept some kind of field that would change the header. For example, in this case I think it would be to also allow some kind of field to `data` that would handle that change.
Check other components for reference. For example: https://github.com/ripe-tech/ripe-components-vue/blob/master/vue/components/ui/atoms/dropdown/dropdown.vue

NFSS10 (changes requested):

NFSS10:
> @3rdvision very good points, perhaps I should have pinned someone before to discuss some important topics about this library 😅
> 
> regarding your questions, yes I have considered reusing the `Form` and `Section` components but it would still require some changes because they are not really the same comparing with the new design. To avoid modifying these components I decided to create a new one but we could try to reuse them if you think is the best idea.

The style shouldn't be a factor when deciding to use a component or not, the components are bland on purpose and you can style them however you want.

Only when you need structural changes is when you should consider expanding the component or maybe, if needed, create a new one.


Also, doesn't ripe-pulse already have this kind of behaviour? If yes see how it's being done and replicate it. I may be wrong though 😅

pmpc94:
@NFSS10 thanks for your inputs. Nope, afaik pulse doesn't have this kind of behaviour as of now.

The way the `form` component is done would raise a lot of complexity when building the object to build the desired component. Considering the following example:
image

We would need to create multiples arrays with multiple columns whereas in the current component we just pass one object with the fields we want to render and then allow its customization whenver the user wants via slots

vue/components/ui/organisms/details-expandable/details-expandable.vue
```diff
@@ -0,0 +1,178 @@
...
+                    <label-ripe
+                        class="details-expandable-title"
+                        v-bind:text="capitalizeName(name)"
+                        v-bind:font-size="labelFontSize"
+                    />
+                </slot>
+                <div class="details-expandable-value">
+                    <slot v-bind:name="sectionName + '-' + name" v-bind:field-value="value">
+                        <input-ripe v-bind:value="value" />
+                    </slot>
```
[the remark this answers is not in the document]
pmpc94: the reason I added a slot here is to give the user an option to customize the input style if needed. What's your suggestion instead?

vue/components/ui/organisms/details-expandable/details-expandable.stories.js
pmpc94: what if we want to change only certain inputs and not all?

NFSS10:
> @NFSS10 thanks for your inputs. Nope, afaik `pulse` doesn't have this kind of behaviour as of now.
> 
> I will change to use the `Section Expandable` component, which is a fairly straightforward change.
> 
> The way the `form` component is done would raise a lot of complexity when building the object to build the desired component. Considering the following example: image
> 
> We would need to create so many arrays just to pass 1 input per column in the first row. Then another column just for the second row. Four columns for the 3rd row etc...
> 
> And then what if we have this: image
> 
> We would need to create multiples arrays with multiple columns whereas in the current component we just pass one object with the fields we want to render and then allow its customization whenever the user wants it via slots


If you think the `form`'s complexity is going to affect very negatively and be too over-engineered for the feature then I would recommend to just create the specific forms and inputs combinations in ripe-pulse and use `section-expandable` for the expandable behaviour.

It will probably be a lot more simple and cleaner doing that way, creating another component will be redundant for the context of ripe-components-vue

pmpc94:
@NFSS10 hmmm that makes sense as well, thanks for your feedback. @joao-conde what do you think? Personally I like this idea to not create just another component in this library and adapt it only to Pulse

joao-conde:
All valid points, seems good to me, I agree with Nuno

Issue #644, opened 2023-11-10
Change radio group padding


pmpc94:
Change radio group padding left and right to 6px (it was 0, making it too close then using e.g., display: flex)

PR #645, opened 2023-11-10
644 - add padding to radio group

pmpc94:
### Issue
- https://github.com/ripe-tech/ripe-components-vue/issues/644


joao-conde (commented):
If you change this in the component it will affect all our UIs using it. Applying the padding directly where you are using it (ripe pulse) is probably for the best.

pmpc94:
applying the changes directly in pulse as this could change other usage of this component

Issue #646, opened 2023-11-13
Add onChange event to input component


pmpc94:
Add onChange event to input component

pmpc94:
not needed, we can use keyup event instead

PR #647, opened 2023-11-13
646 - add onchange event

pmpc94:
### Issue
- https://github.com/ripe-tech/ripe-components-vue/issues/646

### Decisions
- add onChange event


pmpc94:
not needed, we can use keyup event instead

